PR #2, opened 2020-12-13
Refactor 

FuqingWang:
Some update:
1. Refactored code into components for easier module management.
2. switched the app root from index.js to App.js (Optional, but starting with App.js is more like a convention that the react community follows. index.js is more of an application entry point, while app.js is the root of our app tree, where our component hierarchy starts. Generally we don't mess up with the entry point but the app tree)
3. rename the folder from "Components" to "components". (Also a conventional thing, lower case first letter for component folders. But for actual components we use upper case first letter)


mfleader:
I am going to remove the lock files, since we already have `package.json`.

FuqingWang:
Just removed those package-lock.json and yarn.lock. I am looking into where they got added.

PR #3, opened 2020-12-16
adding dockerfile

whitleykeith:
Adds dockerfile that can run the app


frontend/frontend.containerfile
```diff
@@ -1,15 +1,19 @@
-FROM node:14.15.3-alpine3.12 AS build
+# pull official base image
```
mfleader, line 1: The nodejs server is only for development. This image build code should be in frontend-dev.

whitleykeith: we shouldn't need to use nginx because proxying is no longer needed. Realistically we don't even need the two different dockerfiles

mfleader:
looks good to me

PR #9, opened 2021-04-14
landing-page and nav 

whitleykeith:
also some general refactoring such as making the backend call a hook that can be extended for other components.


mfleader:
document airflow service in [application configuration](https://github.com/whitleykeith/ocp_perf_dashboard/pull/1)

PR #54, opened 2023-11-10
remove underlines from list item keys

dustinblack:
## Type of change

- [ ] Refactor
- [ ] New feature
- [ ] Bug fix
- [x] Optimization
- [ ] Documentation Update

## Description

Underlines are confusing in a web interface where they normally imply a hyperlink. Replace the `<u>` tags around the keys in `<li>` items with `<div>` with style references using bold formatting instead.

- Related Issue #
- Closes #

## Testing
I tried to quickly get the yarn environment running for local frontend testing, but it wouldn't work for me and for this simple style change I didn't invest the time to troubleshoot the test environment. I would appreciate it if someone else could validate this real quick.

### DIFF
--- a/backend/backend.containerfile
+++ b/backend/backend.containerfile
@@ -15,7 +15,7 @@ WORKDIR /backend
 
 EXPOSE 8000
 
-RUN dnf install -y pip
+RUN dnf install -y pip gcc python3-devel gcc-c++
 
 RUN pip install --user poetry && \
     poetry export -f requirements.txt -o requirements.txt && \
--- a/frontend/src/components/OCP/InstallCard.js
+++ b/frontend/src/components/OCP/InstallCard.js
@@ -44,13 +44,13 @@ export default function InstallCard(props) {
                     <GridItem rowSpan="1">
                         <Card><CardHeader><CardTitle>Cluster Metadata</CardTitle></CardHeader>
                             <CardBody><ul>
-                                <li><div class="list-item-key">Release Binary</div>: {config.cluster_version  && config.cluster_version || config.ocpVersion}</li>
-                                <li><div class="list-item-key">Cluster Name</div>: {config.cluster_name && config.cluster_name || config.clusterName}</li>
-                                <li><div class="list-item-key">Cluster Type</div>: {config.cluster_type  && config.cluster_type || config.clusterType}</li>
-                                <li><div class="list-item-key">Network Type</div>: {config.network_type  && config.network_type || config.networkType}</li>
-                                <li><div class="list-item-key">Benchmark Status</div>: {icons[config.job_status  && config.job_status || config.jobStatus] || config.job_status  && config.job_status || config.jobStatus}</li>
-                                <li><div class="list-item-key">Duration</div>: {formatTime(config.job_duration  && config.job_duration || config.jobDuration)}</li>
-                                <li><div class="list-item-key">Test ID</div>: {config.uuid}</li>
+                                <li><span class="list-item-key">Release Binary</span>: {config.cluster_version  && config.cluster_version || config.ocpVersion}</li>
+                                <li><span class="list-item-key">Cluster Name</span>: {config.cluster_name && config.cluster_name || config.clusterName}</li>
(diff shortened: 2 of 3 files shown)


chentex:
Hey @dustinblack the change is adding a new line after the div section.

Screenshot from 2023-11-14 12-34-52

jtaleric:
hey all! Is this still relevant?

dustinblack:
Updated the `<div>`s to `<span>`s, which should correct the errant line return.

dustinblack:
The `local-compose.sh` script is failing to build the backend for me, so I still can't validate it locally.

jtaleric:
> @jtaleric I think it's still relevant. The existing underlines are a confusing UXD experience.

ok, it just sat idle for a bit. Thanks

jtaleric:
> error: command 'gcc' failed: No such file or directory

`error: command 'gcc' failed: No such file or directory` 

Do you have gcc?

dustinblack:
> > error: command 'gcc' failed: No such file or directory
> 
> `error: command 'gcc' failed: No such file or directory`
> 
> Do you have gcc?

The `local-compose.sh` script is all `podman` commands, so that error is from the `pip install` within the container build, not from my local system.

dustinblack:
> I wonder if it is as simple as adding the below, or if there is some bigger arch difference here...
> 
> ```
> diff --git a/backend/backend.containerfile b/backend/backend.containerfile
> index 92c0be9..cde57a9 100644
> --- a/backend/backend.containerfile
> +++ b/backend/backend.containerfile
> @@ -15,7 +15,7 @@ WORKDIR /backend
>  
>  EXPOSE 8000
>  
> -RUN dnf install -y pip
> +RUN dnf install -y pip gcc
>  
>  RUN pip install --user poetry && \
>      poetry export -f requirements.txt -o requirements.txt && \
> ```

I've got a few cycles, so I'm testing it now, but `gcc` alone isn't the solution as another python development dependency pops up. Now for some reason my build can't access the repos (outage?), so I'm stuck. I'll give it another shot later.

jtaleric:
> > I wonder if it is as simple as adding the below, or if there is some bigger arch difference here...
> > ```
> > diff --git a/backend/backend.containerfile b/backend/backend.containerfile
> > index 92c0be9..cde57a9 100644
> > --- a/backend/backend.containerfile
> > +++ b/backend/backend.containerfile
> > @@ -15,7 +15,7 @@ WORKDIR /backend
> >  
> >  EXPOSE 8000
> >  
> > -RUN dnf install -y pip
> > +RUN dnf install -y pip gcc
> >  
> >  RUN pip install --user poetry && \
> >      poetry export -f requirements.txt -o requirements.txt && \
> > ```
> 
> I've got a few cycles, so I'm testing it now, but `gcc` alone isn't the solution as another python development dependency pops up. Now for some reason my build can't access the repos (outage?), so I'm stuck. I'll give it another shot later.

ack. Thanks for trying... 

Yeah, I don't have a aarch64 machine to help figure this out. I do know we were able to s/podman/docker (they didn't have podman installed) and have some Mac users get this working (they were on the M1 chipset).

jtaleric:
https://stackoverflow.com/questions/65084318/trouble-installing-pandas-on-new-macbook-air-m1 

I realize you mention you are running linux + arm64 -- I am curious do you have panda's installed via pip on your machine?

jtaleric:
> > https://stackoverflow.com/questions/65084318/trouble-installing-pandas-on-new-macbook-air-m1
> > I realize you mention you are running linux + arm64 -- I am curious do you have panda's installed via pip on your machine?
> 
> I don't, but again, the package installs and dependencies are all within the containers.

ack - yup totally understand that. I as trying to figure out what with the container build was missing due to the arch.



> I fixed the container build for linux/aarch64, but the `local-compose.sh` script still doesn't seem to be enough on its own to stand up a simple local frontend to validate the change.

awesome! Thanks for that. 

Yeah, the toml/ auth for ES is still clunky.

dustinblack:
Bump.

chentex:
Superseeded by #79

PR #59, opened 2023-11-28
Jira Service Connection

mfleader:
## Type of change

- [X] New feature

## Description

Add a connector and http endpoint for Jira so to facilitate the use of Jira data and statistics in a dashboard view.

- Related Issue #
- Closes #

## Testing


backend/app/api/v1/endpoints/cpt/jira.py
chentex, line 1: I am not a 100% convinced this endpoint should be here, but for the moment let's just keep it here, and see how it's use evolves.

chentex (approved):
LGTM

backend/app/api/v1/endpoints/cpt/jira.py
chentex: Yes, that was what I meant, but as I said let's leave it there for the moment, and see how the API evolves to see if we need to move it around.

athiruma:
Why do we need to create this endpoint when Jira already has an [API endpoint](https://developer.atlassian.com/server/jira/platform/jira-rest-api-example-query-issues-6291606/)?
If we need to use it within other endpoints or mappers, we can create a class to access the Jira APIs

mfleader:
@athiruma fair question.

I see it as a trade off between writing and maintaining our own Jira API types and data hooks in JavaScript or TypeScript (or working with another community project like [jira.js](https://github.com/mrrefactoring/jira.js/) or [jira-client](https://github.com/jira-node/node-jira-client)), or adding an endpoint that uses the Atlassian supported [Python API to Jira](https://github.com/atlassian-api/atlassian-python-api).

mfleader:
Somewhat in line with @chentex said, I think as we determine what needs to be queried on the Jira API for the dashboard the trade-off of this endpoint's existence should be re-evaluated. 

At the moment, I do not know what the UI needs from Jira, so this is a flexible solution.

vishnuchalla:
Just curious, why do we need this JIRA connection? Just trying to understand the purpose of it.

chentex (changes requested):
Since I see most of us have doubts about this I propose the following steps:

1. @mfleader has already provided a `Jira Service` that could be used by any teams if they need to query something from `Jira` to their custom pages, that is a win and is done in a way any team could use it eventually.
2. Remove the enpoint, my reasoning:
   1. We do not have plans on displaying raw `Jira` data in the dashboard.
   2. If at some point it is needed to display something, I am sure that data would need to be formatted to a state that the UI could take and display it in a custom UI component, so probably the team doing that would use a specific new endpoint for that part and use the already provided `Jira` service to query their data and then clean it up to display it.

Please let me know if you agree with this @mfleader @athiruma @vishnuchalla

mfleader:
> Since I see most of us have doubts about this I propose the following steps:
> 
> 1. @mfleader has already provided a `Jira Service` that could be used by any teams if they need to query something from `Jira` to their custom pages, that is a win and is done in a way any team could use it eventually.
> 2. Remove the enpoint, my reasoning:
>    
>    1. We do not have plans on displaying raw `Jira` data in the dashboard.
>    2. If at some point it is needed to display something, I am sure that data would need to be formatted to a state that the UI could take and display it in a custom UI component, so probably the team doing that would use a specific new endpoint for that part and use the already provided `Jira` service to query their data and then clean it up to display it.
> 
> Please let me know if you agree with this @mfleader @athiruma @vishnuchalla

By already provided Jira service, do you mean the underlying Jira service that this endpoint is using?

From the conversations outside of this PR, it sounds like folks do want some sort of Jira endpoint here.

I think my point about using [Atlassian's supported Python API](https://github.com/cloud-bulldozer/cpt-dashboard/pull/59#issuecomment-1841504359) still stands.

backend/app/api/v1/endpoints/cpt/jira.py
chentex, line 1: Can you move this to  a new folder called `jira`, inside the `endpoints` folder

backend/app/api/api.py
```diff
@@ -4,18 +4,20 @@
...
-# OCP endopoints
+# OCP endpoints
 router.include_router(ocpJobs.router, tags=['ocp'])
 router.include_router(results.router, tags=['ocp'])
 router.include_router(graph.router, tags=['ocp.graphs'])
 
-# CPT endopoints
+# CPT endpoints
 router.include_router(cptJobs.router, tags=['cpt'])
+router.include_router(jira.router, tags=['cpt.jira'])
```
chentex, line 20: Separate to a new tag, maybe just call it `jira`

backend/app/services/jira_svc.py
```diff
@@ -0,0 +1,24 @@
...
+
+from app import config
+
+
+class JiraService:
+
+    def __init__(self, configpath="jira"):
+        self.cfg = config.get_config()
+        self.url = self.cfg.get(configpath+'.url')
+        self.pat = self.cfg.get(configpath+'.personal_access_token')
```
chentex, line 11: Can you add a section in the README explaining this fields

backend/skeleton.toml
```diff
@@ -7,3 +7,5 @@ password=
 [ocp-server]
 port=8000
 
+[jira]
+host_url=
```
chentex, line 11: This `host_url` doesn't match the one used in the `JiraService` class, and the `personal_access_token` is missing.

chentex (changes requested):
Since the time of my suggestions seems that now there is a need for the endpoint, so scratch all of that,
Seems that the consensus is to provide a JQL to the endpoint to get the needed fields.
i added some comments to reorganize the PR, to merge it after.

chentex (approved):
/lgtm thanks for the changes @mfleader

dustinblack (approved):
LGTM

chentex:
Thanks for the contribution @mfleader !

PR #94, opened 2024-05-03
[Bug fix] Changing if-block scope

vishnuchalla:
## Type of change

- [x] Refactor
- [ ] New feature
- [x] Bug fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Bug fix with if-block that was out of scope earlier. This was resulting in an error in the dashboard saying `new_start_date` variable used without declaration while filtering older time ranges. This PR should fix it. 

## Testing
Tested and verified in local using `sh local-compose.sh`


chentex (approved):
LGTM

PR #126, opened 2024-10-24
Updated pandas version and added numpy package

MVarshini:
## Type of change

- [ ] Refactor
- [ ] New feature
- [x] Bug fix
- [ ] Optimization
- [ ] Documentation Update

## Description

On installing the CPT dashboard python dependencies to start the backend server which was failing due to package version errors. So, got this straightened out by upgrading the pandas package to 2.2.2 and setting numpy to 1.26.4

- Related Issue #
- Closes #

## Testing


dbutenhof (commented):
Moving these forward is probably a good idea (Python 3.9 support ends in about a year). However keep in mind that as long as the CPT dashboard backend runs on Python 3.9, it's also not necessary. The CentOS stream9 container base image used to deploy is 3.9.19.

Also, to fully support Python 3.10, other dependencies (including at least `uvloop`) need to be updated, and I suspect that any such effort should be made independent of any specific CPT dashboard feature.

chentex (approved):

MVarshini:
> Moving these forward is probably a good idea (Python 3.9 support ends in about a year). However keep in mind that as long as the CPT dashboard backend runs on Python 3.9, it's also not necessary. The CentOS stream9 container base image used to deploy is 3.9.19.
> 
> Also, to fully support Python 3.10, other dependencies (including at least `uvloop`) need to be updated, and I suspect that any such effort should be made independent of any specific CPT dashboard feature.

@dbutenhof Got it! As there will be a need to update other dependencies to make them compatible with Python3.10, I'm withdrawing this PR.

PR #147, opened 2024-12-16
Random Pagination ISSUE-143

### LINKED ISSUE #143: REVAMP branch random pagination
Pagination works properly only for `next` page but not when trying to navigate to `previous`, `first` and `last` pages. (i.e. the rows in the table are not getting updated at all). After looking at the logs, this seems like there is no query being sent from frontend to the backend with updated `size/offset` on a state update.

Must be resolved before merging REVAMP to main

cc: @dbutenhof @MVarshini @jaredoconnell @chentex @jtaleric

_Originally posted by @vishnuchalla in https://github.com/cloud-bulldozer/cpt-dashboard/issues/135#issuecomment-2541094304_

MVarshini:
Random Pagination not working

## Type of change

- [ ] Refactor
- [ ] New feature
- [x] Bug fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Pagination on clicking `first` and `last` buttons not working

- Related Issue #
- Closes #143 

## Testing

### DIFF
--- a/backend/app/api/v1/endpoints/cpt/cptJobs.py
+++ b/backend/app/api/v1/endpoints/cpt/cptJobs.py
@@ -13,6 +13,7 @@ from .maps.telco import telcoMapper
 from .maps.ocm import ocmMapper
 from ...commons.example_responses import cpt_200_response, response_422
 from fastapi.param_functions import Query
+from app.api.v1.commons.utils import normalize_pagination
 
 router = APIRouter()
 
@@ -68,13 +69,7 @@ async def jobs(
             status_code=422,
         )
 
-    if offset and not size:
-        raise HTTPException(400, f"offset {offset} specified without size")
-    elif not offset and not size:
-        size = 10000
-        offset = 0
-    elif not offset:
-        offset = 0
+    offset, size = normalize_pagination(offset, size)
 
     results_df = pd.DataFrame()
     total_dict = {}
@@ -110,7 +105,7 @@ async def jobs(
         "endDate": end_date.__str__(),
         "results": results_df.to_dict("records"),
         "total": jobsCount,
-        "offset": (offset + size) if size != 10000 else 0,
+        "offset": offset + size,
     }
 
     if pretty:
--- a/backend/app/api/v1/endpoints/cpt/cptJobs.py
+++ b/backend/app/api/v1/endpoints/cpt/cptJobs.py
@@ -2,7 +2,7 @@ import json
 import asyncio
 from concurrent.futures import ProcessPoolExecutor, as_completed
 from multiprocessing import cpu_count
-from fastapi import Response
+from fastapi import Response, HTTPException
 import pandas as pd
 from datetime import datetime, timedelta, date
 from fastapi import APIRouter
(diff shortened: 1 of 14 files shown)


backend/app/api/v1/endpoints/cpt/cptJobs.py
```diff
@@ -102,7 +102,7 @@ async def jobs(
         "endDate": end_date.__str__(),
         "results": results_df.to_dict("records"),
         "total": jobsCount,
-        "offset": offset + size,
+        "offset": offset + size if size else 0,
```
dbutenhof, line 105: In some APIs (and you added some more with this PR), you're protecting against having `offset` set to a value without `size`. That's not the case here, and this expression isn't considering the value of `offset`. So `if size` we add `offset + size` (regardless of `offset`) or else use `0`.

Is that what you meant? Should you be raising the `HTTPException` here `if offset and not size`?

```python
>>> offset = 10
>>> size = None
>>> offset + size if size else 0
0
>>> offset + (size if size else 0)
10
```

backend/app/api/v1/endpoints/cpt/cptJobs.py
```diff
@@ -102,7 +102,7 @@ async def jobs(
         "endDate": end_date.__str__(),
         "results": results_df.to_dict("records"),
         "total": jobsCount,
-        "offset": offset + size,
+        "offset": (offset + size) if size != 10000 else 0,
```
dbutenhof, line 105: Where does the `10000` come from here? In some of the other APIs, you *set* it if the query parameter is 0, but not here. If the call is made without `?size=<n>`, won't size be `None`?

backend/app/api/v1/endpoints/telco/telcoJobs.py
```diff
@@ -60,8 +67,8 @@ async def jobs(
         "startDate": start_date.__str__(),
         "endDate": end_date.__str__(),
         "results": jobs,
-        "total": 0,
-        "offset": 0,
+        "total": results["total"],
+        "offset": offset + size if size else 0,
```
dbutenhof, line 71: Is it intentional that this differs from the `ocpJobs` logic above, where we're checking for `size != 10000` rather than just `if size`? If `offset` is 0, and `size` isn't specified, that one will return 0 and this will return 10000.

backend/app/api/v1/endpoints/cpt/cptJobs.py
jaredoconnell: Yeah, without context this seems odd.

jaredoconnell (commented):
My concerns match Dave's.

backend/app/api/v1/endpoints/cpt/cptJobs.py
MVarshini: Yeah, that's right. I missed to check for `offset` and `size` in the query parameters. Updated the code.

backend/app/api/v1/endpoints/telco/telcoJobs.py
MVarshini: It slipped my mind to add the `size != 10000` here. Thanks for catching it!

dbutenhof (approved):
Looks good, except that you're running into the same Poetry 2.0 problems that Vishnu hit. He needs to fix his fix, and you'll need to merge or copy that onto your `revamp` branch.

(Sigh. If only the world would stand still while we aim ... 😆 )

jaredoconnell (approved):
It's nice to see the normalization code moved to a function.

PR #157, opened 2025-01-23
Applying Filters to the Telco Table

MVarshini:
## Type of change

- [ ] Refactor
- [ ] New feature
- [ ] Bug fix
- [ ] Optimization
- [ ] Documentation Update

## Description

[PANDA-700](https://issues.redhat.com/browse/PANDA-700)

This PR filters the data of Telco table 

- Related Issue #
- Closes #

## Testing


backend/app/api/v1/commons/telco.py
```diff
@@ -112,13 +130,21 @@ async def getFilterData(start_datetime: date, end_datetime: date, configpath: st
...
     )
+    searchList = test_type_filter
+    if filter:
+        filter_dict = utils.get_dict_from_qs(filter)
+        searchQuery = utils.construct_query(filter_dict)
+        if "benchmark" in filter_dict:
+            searchList = searchQuery
+        else:
+            searchList = searchQuery + " " + test_type_filter
+
```
dbutenhof, line 144: This seems to be repeated between `getFilterData` and `getData` ... can it be factored out into a helper function?

backend/app/api/v1/commons/utils.py
```diff
@@ -133,3 +134,25 @@ def buildReleaseStreamFilter(input_array):
...
+
+
+def construct_query(filter_dict):
+    query_parts = []
+    if isinstance(filter_dict, dict):
+        for key, values in filter_dict.items():
+            k = constants.FIELDS_FILTER_DICT[key]
+            if len(values) > 1:
+                or_clause = " or ".join([f'{k}="{value}"' for value in values])
+                query_parts.append(f"{or_clause}")
```
dbutenhof, line 155: This is really just `query_parts.append(or_clause)` 😆

backend/app/api/v1/commons/utils.py
```diff
@@ -133,3 +134,25 @@ def buildReleaseStreamFilter(input_array):
...
+    return cleaned_dict
+
+
+def construct_query(filter_dict):
+    query_parts = []
+    if isinstance(filter_dict, dict):
+        for key, values in filter_dict.items():
+            k = constants.FIELDS_FILTER_DICT[key]
+            if len(values) > 1:
+                or_clause = " or ".join([f'{k}="{value}"' for value in values])
```
dbutenhof, line 154: Elsewhere you capitalized the Splunk keyword `OR`. It'd be a bit easier to read if you're consistent about such conventions. The all-caps *does* make it stand out, and is probably worthwhile.

backend/app/services/splunk.py
```diff
@@ -116,24 +111,15 @@ async def filterPost(self, query, searchList=""):
...
             else:
-                search_query = "search index={} | stats count AS total_records, values(cpu) AS cpu, values(node_name) AS nodeName, values(test_type) AS benchmark, values(ocp_version) AS ocpVersion, values(ocp_build) AS releaseStream".format(
-                    self.indice
-                )
+                search_query = f"search index={self.indice} | stats count AS total_records, values(cpu) AS cpu, values(node_name) AS nodeName, values(test_type) AS benchmark, values(ocp_version) AS ocpVersion, values(ocp_build) AS releaseStream"
 
             try:
                 # Run the job and retrieve results
-                job = await self.service.jobs.create(
+                job = self.service.jobs.create(
```
dbutenhof, line 122: Above where you removed the `await`, you replaced it with a `not job.is_done()` busy loop. I think we discussed a similar change earlier, though I don't remember exactly why it was better ... but more importantly you don't seem to be doing anything here to wait for `job` before referencing the results at line 134 ... won't that fail?

backend/app/api/v1/commons/telco.py
```diff
@@ -40,58 +45,71 @@ async def getData(
         "latest_time": "{}T23:59:59".format(end_datetime.strftime("%Y-%m-%d")),
         "output_mode": "json",
     }
-    searchList = " OR ".join(
+    test_type_filter = " OR ".join(
```
dbutenhof, line 48: I shouldn't be commenting on lines you didn't change, but in other places you've changed the awkward old `str.format()` substitutions to more readable and modern f-strings -- that'd look better here, too. 😁

backend/app/api/v1/commons/telco.py
```diff
@@ -40,58 +45,71 @@ async def getData(
...
+                    + "/"
+                    + str(test_data["cluster_artifacts"]["ref"]["jenkins_build"]),
+                    "jobStatus": "failure" if (threshold != 0) else "success",
+                    "jobDuration": execution_time_seconds,
+                }
+            )
 
     jobs = pd.json_normalize(mapped_list)
 
     return {"data": jobs, "total": response["total"]}
```
jaredoconnell, line 107: You added an `if response` check, but you then use the response value afterwards, not conditioned by it. What situation does this happen, and why doesn't it also apply to the "total" key/value?
If you want to do something else in that situation, you can add a condition for the not response, and exit early.

MVarshini: Added the check for `total` too!

backend/app/api/v1/commons/telco.py
MVarshini: Modified it :)

backend/app/api/v1/commons/utils.py
MVarshini: Gotcha!

backend/app/services/splunk.py
MVarshini: Yeah reverted the change as `await` in that line is throwing error.

dbutenhof: Ah, OK -- because the `splunklib` `Jobs.create()` doesn't return a Python `Coroutine` at all, so `await` won't work. But the semantics of the Splunk `Job.results()` isn't entirely clear in the documentation I saw. Is the `while not job.is_done()` loop above necessary? And, if so, don't you also need that here? `Job.results()` returns a "stream", so it's quite possible that the explicit wait loop *isn't* necessary, and we'll just block at the read until there are results, reading them one-by-one. And if that works here, would it work up above at line 66/67 as well instead of the explicit wait loop???

backend/app/api/v1/commons/utils.py
dbutenhof: You didn't change this? On the one hand, it won't break anything, but it's an awkward usage that'll unnecessarily draw the attention of a reader away from the code flow. 😆

dbutenhof (commented):
I suspect that this is fine, although I'm a little concerned about the explicit vs implicit wait patterns for the splunk jobs (unless there's a real reason, doing it the same both places would be better), and I'd prefer getting rid of the unnecessary f-string.

backend/app/api/v1/commons/utils.py
MVarshini: It got missed. Now, updated it :grin:

backend/app/services/splunk.py
MVarshini: If there is no wait loop, then oneshot is not streaming the results, causing the query to return only the `total_records` value.

backend/app/services/splunk.py
```diff
@@ -111,11 +111,19 @@ async def filterPost(self, query, searchList=""):
...
-            search_query = ""
+            search_query = f"search index={self.indice} "
+
             if searchList:
-                search_query = f"search index={self.indice} {searchList} | stats count AS total_records, values(cpu) AS cpu, values(node_name) AS nodeName, values(test_type) AS benchmark, values(ocp_version) AS ocpVersion, values(ocp_build) AS releaseStream"
-            else:
-                search_query = f"search index={self.indice} | stats count AS total_records, values(cpu) AS cpu, values(node_name) AS nodeName, values(test_type) AS benchmark, values(ocp_version) AS ocpVersion, values(ocp_build) AS releaseStream"
+                search_query += f"{searchList} | "
+
+            search_query += (
```
dbutenhof, line 119: I like that you removed the redundancy here!

But ... the "not `searchList`" case also had a `|` after the index phrase (`search index={self.indice} |`), and the new version doesn't. Is that correct?

backend/app/services/splunk.py
dbutenhof: I think maybe that makes sense to me. Although I'm not sure whether I need to learn more about Splunk or whether I want to run and hide. 😆 

So the wait loop above is because you're using the `oneshot` to stream results, whereas here you're just iterating on `job.results` directly. Huh. I'll take your word for it, but thanks for the patient explanation. 😆

dbutenhof (commented):
I think you've resolved all my previous comments, but I have a new question on your Splunk query string refactoring ...

backend/app/services/splunk.py
dbutenhof: So this now becomes `search index={self.indice} | {searchList} |`, with *two* pipes instead of one. I don't know the Splunk syntax -- is that OK?

dbutenhof (approved):

dbutenhof:
@vishnuchalla @chentex @jtaleric @jaredoconnell I think this one's looking good, and I approved ... but I'm not seeing any other feedback. It'd be nice to get another eye or two on this, but at some point I'll just merge it so we can move this ahead. Any takers!? 😏

jaredoconnell (approved):

PR #160, opened 2025-02-10
Adding backend for openshift lightspeed

vishnuchalla:
## Type of change

- [ ] Refactor
- [x] New feature
- [ ] Bug fix
- [ ] Optimization
- [ ] Documentation Update

## Description
Adding backend code changes for openshift lightspeed integration.

Closes: https://issues.redhat.com/browse/PERFSCALE-3519

## Testing
Tested using `sh local-compose.sh`. Verified both listing and graph for runs.
Graph endpoint results
```
{
  "latencyResults": [
    {
      "x": [
        "get_feedback_status",
        "get_liveness",
        "get_metrics",
        "get_readiness",
        "post_authorized",
        "post_feedback",
        "post_query",
        "post_query_with_cache",
        "post_streaming_query",
        "post_streaming_query_with_cache"
      ],
      "y": [
        412839.651,
        426321.545,
        1257025.449,
        428960.542,
        962199.639,
        1219118.85,
        7515738.922,
        7722044.497,
        7746687.88,
        7693166.952
      ],
      "name": "Previous API Latencies",
      "type": "bar",
      "orientation": "v"
    },
    {
      "x": [
        "get_feedback_status",
        "get_liveness",
        "get_metrics",
        "get_readiness",
        "post_authorized",
        "post_feedback",
        "post_query",
        "post_query_with_cache",
        "post_streaming_query",
        "post_streaming_query_with_cache"
      ],
      "y": [
        426681.054,
        442130.035,
        1321660.427,
        443547.009,
        1065190.981,
        1311414.425,
        7640858.78,
        7533090.526,
        7917279.162,
        7916161.262
      ],
      "name": "Current API Latencies",
      "type": "bar",
      "orientation": "v"
    }
  ],
  "apiResults": [
    {
      "x": [
        "get_feedback_status",
        "get_liveness",
        "get_metrics",
        "get_readiness",
        "post_authorized",
        "post_feedback",
        "post_query",
        "post_query_with_cache",
        "post_streaming_query",
        "post_streaming_query_with_cache"
      ],
      "y": [
        233627,
        228881,
        35775,
        225272,
        50936,
        37453,
        4818,
        4739,
        4364,
        4319
      ],
      "name": "Previous API Throughput",
      "type": "bar",
      "orientation": "v"
    },
    {
      "x": [
        "get_feedback_status",
        "get_liveness",
        "get_metrics",
        "get_readiness",
        "post_authorized",
        "post_feedback",
        "post_query",
        "post_query_with_cache",
        "post_streaming_query",
        "post_streaming_query_with_cache"
      ],
      "y": [
        225555,
        220071,
        30974,
        218676,
        48856,
        32270,
        4789,
        4722,
        4199,
        4200
      ],
      "name": "Current API Throughput",
      "type": "bar",
      "orientation": "v"
    }
  ]
}
```


dbutenhof (approved):
Not knowing the exact schema of the data it's hard to fully evaluate the logic; but since you've presumably tested and validated this, I can at least say the code looks "plausible" and more importantly shouldn't break anything else!

(I'm also hoping that it won't be a pain for Varshini to merge into REVAMP...)

vishnuchalla:
> Not knowing the exact schema of the data it's hard to fully evaluate the logic; but since you've presumably tested and validated this, I can at least say the code looks "plausible" and more importantly shouldn't break anything else!
> 
> (I'm also hoping that it won't be a pain for Varshini to merge into REVAMP...)

As per my discussion with @MVarshini offline there seems to be one final PR to merge revamp into main. Once that is done, I can rebase mine on top of that.

backend/app/api/v1/endpoints/ols/olsJobs.py
```diff
@@ -0,0 +1,78 @@
...
+    pretty: bool = Query(False, description="Output content in pretty format."),
+    size: int = Query(None, description="Number of jobs to fetch"),
+    offset: int = Query(None, description="Offset Number to fetch jobs from"),
+    sort: str = Query(None, description="To sort fields on specified direction"),
+    filter: str = Query(None, description="Query to filter the jobs"),
+):
+
+    return await ocpJobs(start_date, end_date, 
+                         pretty, size, offset, sort, 
+                         filter="benchmark='ols-load-generator'")
```
dbutenhof, line 48: This ignores any *incoming* filter query parameter, so the OLS tab won't actually support filtering within OLS runs. I think you need to *add* this to an existing filter string, and *replace* it only if it's `None`.

backend/app/api/v1/endpoints/cpt/maps/ocp.py
```diff
@@ -35,6 +35,7 @@ async def ocpMapper(
         return {"data": df, "total": response.get("total", 0)}
 
     df.insert(len(df.columns), "product", "ocp")
+    df.loc[df["benchmark"] == "ols-load-generator", "product"] = "ols"
```
dbutenhof, line 38: I copied this for my OCPv prototype, but I've questioned it internally several times.

It has the same potentially bad interaction with backend pagination as the opposite case on the OLS API -- if the returned Opensearch "page" has 5 OLS runs, you're dropping them and the page is 5 short even though there's a next page. At best, that's kinda ugly.

We also don't have "not" filter syntax, so we can't currently add a `benchmark!='ols-load-generator'` filter to compensate on the query.

I think there are two practical solutions -- one is to simply not worry about it, and let `ols-load-generator` and `virt-density` runs show up on the OCP page. The other, which I think is probably much better, especially if/when OLS or OCPv add additional benchmarks (which I know OCPv plans to do) would be for both projects to move out of the default `perf_scale_ci` index and into their own indices (even if that's just `indice = perf_scale_ci` + `prefix = ols_` on the same server[s]) so that we can drop the hardcoded filters and not worry about the OCP queries...

I've recommended that solution to OCPv ... though I haven't gotten any response.

dbutenhof (commented):
General question ... I noticed that you've dropped several additional APIs from OCP in both the graphs and jobs modules. None of these are called by the UI ... does OCP actually call them from some other client which isn't relevant to OLS, or were they always just "dead code"??

backend/app/api/v1/endpoints/ols/olsJobs.py
vishnuchalla: Thanks for bringing this up. Updated logic according to the suggestion.

backend/app/api/v1/endpoints/cpt/maps/ocp.py
vishnuchalla: I have already considered the above concerns. Since OLS is an openshift's layered product, I am going with approach 1 of having it to be appeared on the OCP page.

As far line#38 that has been highlighted above, it doesn't have anything to do with your concerns. I doesn't drop any data, it just transforming the product column to populate the `ols` as value for the runs that have `ols-load-generator` as a benchmark. Its just a mutation which is appropriate for the data here in our case.

vishnuchalla:
> General question ... I noticed that you've dropped several additional APIs from OCP in both the graphs and jobs modules. None of these are called by the UI ... does OCP actually call them from some other client which isn't relevant to OLS, or were they always just "dead code"??

I am not sure what you are referring to here. I am not dropping any existing functionality, just adding on top of it. Please help me understand If I am missing something.

dbutenhof:
> > General question ... I noticed that you've dropped several additional APIs from OCP in both the graphs and jobs modules. None of these are called by the UI ... does OCP actually call them from some other client which isn't relevant to OLS, or were they always just "dead code"??
> 
> I am not sure what you are referring to here. I am not dropping any existing functionality, just adding on top of it. Please help me understand If I am missing something.

And I thought I was just asking you to help *me* understand if I was missing something!

Under the OCP endpoint, I see

`graph.py`:
* '/api/v1/ocp/graph/trend/{version}/{count}/{benchmark}'
* '/api/v1/ocp/graph/trend/{version}/{prev_version}/{count}/{benchmark}/cpu/{namespace}'
* '/api/v1/ocp/graph/trend/{version}/{count}/{benchmark}/cpu/{namespace}'

`result.py`:
* '/api/v1/ocp/jobs/{ci}/{job_id}'

None of these are used in the UI, and you haven't adapted them for OLS. I was just wondering why. Were they experimental APIs that were never used in the first place (or are no longer being used)? Or are they used somewhere outside the UI that isn't relevant for OLS?

backend/app/api/v1/endpoints/ols/olsJobs.py
```diff
@@ -0,0 +1,80 @@
...
+    pretty: bool = Query(False, description="Output content in pretty format."),
+    size: int = Query(None, description="Number of jobs to fetch"),
+    offset: int = Query(None, description="Offset Number to fetch jobs from"),
+    sort: str = Query(None, description="To sort fields on specified direction"),
+    filter: str = Query(None, description="Query to filter the jobs"),
+):
+    if not filter:
+        filter = "benchmark='ols-load-generator'"
+    elif "benchmark='ols-load-generator'" not in filter:
+        filter += "&benchmark='ols-load-generator'"
```
dbutenhof, line 48: When I was playing around with OCPv I avoided the string duplication like this, but it's not a big deal:

```python
    bf = "benchmark=virt-density"
    f = f"{filter}&{bf}" if filter else bf
```

dbutenhof (dismissed):

backend/app/api/v1/endpoints/ols/olsJobs.py
vishnuchalla: good suggestion, Dave. Let me update accordingly real quick.

vishnuchalla:
> > > General question ... I noticed that you've dropped several additional APIs from OCP in both the graphs and jobs modules. None of these are called by the UI ... does OCP actually call them from some other client which isn't relevant to OLS, or were they always just "dead code"??
> > 
> > 
> > I am not sure what you are referring to here. I am not dropping any existing functionality, just adding on top of it. Please help me understand If I am missing something.
> 
> And I thought I was just asking you to help _me_ understand if I was missing something!
> 
> Under the OCP endpoint, I see
> 
> `graph.py`:
> 
> * '/api/v1/ocp/graph/trend/{version}/{count}/{benchmark}'
> * '/api/v1/ocp/graph/trend/{version}/{prev_version}/{count}/{benchmark}/cpu/{namespace}'
> * '/api/v1/ocp/graph/trend/{version}/{count}/{benchmark}/cpu/{namespace}'
> 
> `result.py`:
> 
> * '/api/v1/ocp/jobs/{ci}/{job_id}'
> 
> None of these are used in the UI, and you haven't adapted them for OLS. I was just wondering why. Were they experimental APIs that were never used in the first place (or are no longer being used)? Or are they used somewhere outside the UI that isn't relevant for OLS?

Ohh I see. Yes all those methods were experimental and are dead code. We can get them removed.

vishnuchalla:
@dbutenhof because I update the PR, looks like it needs approval one more time.

backend/app/api/v1/endpoints/ols/olsJobs.py
```diff
@@ -42,10 +42,16 @@ async def jobs(
...
-
-    return await ocpJobs(start_date, end_date, 
-                         pretty, size, offset, sort, 
-                         filter="benchmark='ols-load-generator'")
+    filter_clause = "benchmark='ols-load-generator'"
+    filter = (
+        filter_clause
+        if not filter
+        else filter
+        if filter_clause in filter
```
dbutenhof, line 50: I briefly considered worrying about that case and decided not to; but now you're making me reconsider. On the other hand, while I believe the standard UI action will indeed use exactly `benchmark='ols-load-generator'`, I don't think that `TableFilter` will actually manifest a filter option if `ols-load-generator` is the only `benchmark` value. And the filter parser would generate the same match JSON with `benchmark=ols-load-generator` or `benchmark="ols-load-generator"` or `benchmark=['ols-load-generator']` and I really didn't want to deal with that.

So this attempted query optimization probably doesn't hurt, but might not really help much, either... and my way kept it all on one line. 😁

dbutenhof (approved):

dbutenhof:
> @dbutenhof because I updated the PR, looks like it needs approval one more time.

Approved ... with a pre-resolved comment.

backend/app/api/v1/endpoints/ols/olsJobs.py
vishnuchalla: yeah, just thought if I am doing it let's get it done completely.

PR #163, opened 2025-02-12
Improve and document the TableFilter component

jaredoconnell:
I found the TableFilter component confusing due to the duplicate input and lack of documentation. This PR aims to fix that.

- Adds a notice when no filter options are present, or when no options are present for that filter.
- Reduces duplicate inputs by utilizing existing input to get necessary information
- Documents TableFilters and ColumnMenuFilter components

This was entirely undocumented before, so let me know if I have any documentation errors.

image

## Type of change

- [x] Refactor
- [ ] New feature
- [ ] Bug fix
- [ ] Optimization
- [x] Documentation Update

- Related Issue #
- Closes #

## Testing

### DIFF
--- a/frontend/src/components/molecules/ColumnMenuFilter/index.jsx
+++ b/frontend/src/components/molecules/ColumnMenuFilter/index.jsx
@@ -19,7 +19,8 @@ import { useSelector } from "react-redux";
  * Selection changes are communicated through the callback.
  * 
  * @param {*} props
- *  - type: The project ID type within the CPT Dashboard.
+ *  - type: The project ID type within the CPT Dashboard that as used in
+ *      the commonActions action file.
  *  - setColumns: A callback with two inputs, `value` and `isAdding`:
  *   - value: The column to modify
  *   - isAdding: `true` if adding the column, `false` if removing it.
--- a/frontend/src/components/molecules/ColumnMenuFilter/index.jsx
+++ b/frontend/src/components/molecules/ColumnMenuFilter/index.jsx
@@ -14,11 +14,15 @@ import { uid } from "@/utils/helper";
 import { useSelector } from "react-redux";
 
 /**
+ * A component that displays a list of columns available for the project
+ * ID type, and allows the user to select or deselect them.
+ * Selection changes are communicated through the callback.
+ * 
  * @param {*} props
  *  - type: The project ID type within the CPT Dashboard.
- *  - setColumns: A callback with two inputs, value and isAdding:
+ *  - setColumns: A callback with two inputs, `value` and `isAdding`:
  *   - value: The column to modify
- *   - isAdding: True if adding the column, False if removing it.
+ *   - isAdding: `true` if adding the column, `false` if removing it.
  */
 const ColumnMenuFilter = (props) => {
   const [isOpen, setIsOpen] = React.useState(false);
--- a/frontend/src/components/molecules/ColumnMenuFilter/index.jsx
+++ b/frontend/src/components/molecules/ColumnMenuFilter/index.jsx
@@ -19,7 +19,7 @@ import { useSelector } from "react-redux";
  * Selection changes are communicated through the callback.
  * 
  * @param {*} props
- *  - type: The project ID type within the CPT Dashboard that as used in
+ *  - type: The project ID type within the CPT Dashboard as used in
  *      the commonActions action file.
  *  - setColumns: A callback with two inputs, `value` and `isAdding`:
  *   - value: The column to modify
--- a/frontend/src/components/molecules/ColumnMenuFilter/index.jsx
+++ b/frontend/src/components/molecules/ColumnMenuFilter/index.jsx
@@ -13,6 +13,13 @@ import TableColumnIcon from "@/components/atoms/TableColumnIcon";
 import { uid } from "@/utils/helper";
 import { useSelector } from "react-redux";
 
+/**
+ * @param {*} props
+ *  - type: The project ID type within the CPT Dashboard.
+ *  - setColumns: A callback with two inputs, value and isAdding:
+ *   - value: The column to modify
+ *   - isAdding: True if adding the column, False if removing it.
+ */
 const ColumnMenuFilter = (props) => {
   const [isOpen, setIsOpen] = React.useState(false);
 
@@ -72,6 +79,7 @@ const ColumnMenuFilter = (props) => {
     </Select>
   );
 };
+
 
 ColumnMenuFilter.propTypes = {
   type: PropTypes.string,
--- a/frontend/src/components/molecules/MultiSelectBox/index.jsx
+++ b/frontend/src/components/molecules/MultiSelectBox/index.jsx
@@ -24,7 +24,7 @@ import { useState } from "react";
  *   - `name`: The name of the category.
  *   - `value`: An array of all selected values as strings.
  *  - applyMethod: A callback that is called when it is time to apply the changes (like when
- *      the multi-select is closed).
+ *      the multi-select is closed). Takes no parameters.
  *  - currCategory: A string specifying the current category.
  *  - width: the CSS value for the width of the outer sub-component.
  */
--- a/frontend/src/components/molecules/MultiSelectBox/index.jsx
+++ b/frontend/src/components/molecules/MultiSelectBox/index.jsx
@@ -12,6 +12,22 @@ import {
 import PropTypes from "prop-types";
 import { useState } from "react";
 
+/**
+ * A multi-select box that allows selecting values for a category.
+ *
+ * @param {*} props:
+ *  - options: an array of options that can be selected.
+ *  - onChange: a callback function that is called with inputs `category` and `value`
+ *   - `category`: The value from currCategory.
+ *   - `value`: The value that was selected.
+ *  - selected: The currently selected options in the form of an object with fields `name` and `value`
+ *   - `name`: The name of the category.
+ *   - `value`: An array of all selected values as strings.
+ *  - applyMethod: A callback that is called when it is time to apply the changes (like when
+ *      the multi-select is closed).
+ *  - currCategory: A string specifying the current category.
+ *  - width: the CSS value for the width of the outer sub-component.
+ */
 const MultiSelectBox = (props) => {
   const [isOpen, setIsOpen] = useState(false);
   const [isDirty, setIsDirty] = useState(false);
@@ -31,6 +47,8 @@ const MultiSelectBox = (props) => {
     }
     setIsDirty(false);
   };
+  const message = props.options.length > 0 ? "Select a value" : "No values present";
+
   const toggle = (toggleRef) => {
     return (
       <MenuToggle
@@ -66,7 +84,7 @@ const MultiSelectBox = (props) => {
           </TextInputGroup>
         ) : (
           <TextInputGroup>
-            <TextInputGroupMain value={"Select a value"} />
+            <TextInputGroupMain value={message} />
           </TextInputGroup>
         )}
       </MenuToggle>
--- a/frontend/src/components/organisms/TableFilters/index.jsx
+++ b/frontend/src/components/organisms/TableFilters/index.jsx
@@ -3,6 +3,7 @@ import "react-calendar/dist/Calendar.css";
 import "./index.less";
 
 import {
+  Banner,
   Chip,
   ChipGroup,
   Toolbar,
@@ -25,7 +26,9 @@ import SelectBox from "@/components/molecules/SelectBox";
 import { formatDate } from "@/utils/helper";
 
 /**
- * A component that provides a filter selector, a date selector, and navigation components.
+ * A component that provides an all-in-one toolbar for the tables in this project.
+ * 
+ * Includes a filter selector, a date selector, a column selector, and navigation components.
  * This component utilizes filterActions to modify the data store.
  *
  * @param {*} props:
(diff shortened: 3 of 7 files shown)


frontend/src/components/molecules/ColumnMenuFilter/index.jsx
```diff
@@ -13,6 +13,13 @@ import TableColumnIcon from "@/components/atoms/TableColumnIcon";
 import { uid } from "@/utils/helper";
 import { useSelector } from "react-redux";
 
+/**
+ * @param {*} props
```
dbutenhof, line 17: How about a paragraph describing what the component *does*, as you did elsewhere? (E.g., `TableFilters`.)

frontend/src/components/organisms/TableFilters/index.jsx
```diff
@@ -41,10 +64,22 @@ const TableFilter = (props) => {
...
   } = props;
 
-  const category =
-    filterData?.length > 0 &&
-    categoryFilterValue &&
-    filterData.filter((item) => item.name === categoryFilterValue)?.[0]?.key;
+
+  tableFilters.forEach(function(entry) {
+    console.log(entry);
+  });
```
dbutenhof, line 70: Probably don't leave this in?

(And, FYI, using a `function` for the callback really looks odd since we use lambdas everywhere else.)

frontend/src/components/organisms/TableFilters/index.jsx
```diff
@@ -53,6 +88,8 @@ const TableFilter = (props) => {
     return filter.name;
   };
 
+  const category = getFilterID(categoryFilterValue);
```
dbutenhof, line 91: Do we really want `category` to be "no-filters" or ""? Your logic below doesn't render the multi-select box if `category` is `false`, which seems to correspond to the old logic ... your new `category` won't ever be "false"...

frontend/src/components/organisms/TableFilters/index.jsx
```diff
@@ -41,10 +64,22 @@ const TableFilter = (props) => {
...
+  tableFilters.forEach(function(entry) {
+    console.log(entry);
+  });
+
+
+  const getFilterID = (name) => {
+    if (!(tableFilters?.length > 0)) {
+      return "no-filters";
+    }
+    const filterResults = tableFilters.filter((item) => item.name === name);
```
dbutenhof, line 77: Without reading all the action and reducer code ... is this really saying that the previous `filterData` is just the same as `tableFilters`??

frontend/src/components/organisms/TableFilters/index.jsx
```diff
@@ -72,28 +109,37 @@ const TableFilter = (props) => {
...
+                <SelectBox
+                  options={tableFilters}
+                  onChange={onCategoryChange}
+                  selected={categoryFilterValue}
+                  icon={<FilterIcon />}
+                  width={"200px"}
+                />
+              </ToolbarItem>
+              <ToolbarItem>
+                {category != false ? (
```
dbutenhof, line 125: `{category ? ( ... )}`?? But while `""` might be falsey in Javascript, `"no-filter"` definitely isn't. If this is what you want, a comment would help...

frontend/src/components/organisms/TableFilters/index.jsx
```diff
@@ -24,12 +24,35 @@ import PropTypes from "prop-types";
 import SelectBox from "@/components/molecules/SelectBox";
 import { formatDate } from "@/utils/helper";
 
+/**
+ * A component that provides a filter selector, a date selector, and navigation components.
+ * This component utilizes filterActions to modify the data store.
```
dbutenhof, line 29: This is a complicated widget that I think could use a bit more explanation.

dbutenhof (commented):
First pass comments ... I may need to try to grok the component actions and reducers to really understand exactly what you've done by removing the `filterData` prop...

frontend/src/components/organisms/TableFilters/index.jsx
jaredoconnell: I made some improvements to the description. If you don't think it's enough, what extra points do you think I should cover?

jaredoconnell (commented):
Addressed review comments. Also, I made improvements to MultiSelectBox.

frontend/src/components/molecules/ColumnMenuFilter/index.jsx
```diff
@@ -14,11 +14,16 @@ import { uid } from "@/utils/helper";
...
 
 /**
+ * A component that displays a list of columns available for the project
+ * ID type, and allows the user to select or deselect them.
+ * Selection changes are communicated through the callback.
+ * 
  * @param {*} props
- *  - type: The project ID type within the CPT Dashboard.
- *  - setColumns: A callback with two inputs, value and isAdding:
+ *  - type: The project ID type within the CPT Dashboard that as used in
```
dbutenhof, line 22: "that *is* used"? Or should the sentence be structured differently?

frontend/src/components/molecules/MultiSelectBox/index.jsx
```diff
@@ -12,6 +12,22 @@ import {
...
+ *
+ * @param {*} props:
+ *  - options: an array of options that can be selected.
+ *  - onChange: a callback function that is called with inputs `category` and `value`
+ *   - `category`: The value from currCategory.
+ *   - `value`: The value that was selected.
+ *  - selected: The currently selected options in the form of an object with fields `name` and `value`
+ *   - `name`: The name of the category.
+ *   - `value`: An array of all selected values as strings.
+ *  - applyMethod: A callback that is called when it is time to apply the changes (like when
```
dbutenhof, line 26: Not a problem ... but I skimmed down to find the call to be sure there weren't any parameters. There aren't, so this is correct ... but maybe saying ("no params") wouldn't be a bad idea to save others the effort?

frontend/src/components/organisms/TableFilters/index.jsx
dbutenhof: That looks *lots* better! Thanks. 🥳

frontend/src/components/molecules/MultiSelectBox/index.jsx
jaredoconnell: I added a note that it takes no parameters.

frontend/src/components/molecules/ColumnMenuFilter/index.jsx
jaredoconnell: That looks like it ended up as a mix of "is is used in" and "as used in". I switched it to "as used in".

frontend/src/components/organisms/TableFilters/index.jsx
```diff
@@ -41,10 +67,16 @@ const TableFilter = (props) => {
     updateSelectedFilter,
   } = props;
 
-  const category =
-    filterData?.length > 0 &&
-    categoryFilterValue &&
-    filterData.filter((item) => item.name === categoryFilterValue)?.[0]?.key;
+  const getFilterID = (name) => {
```
MVarshini, line 70: Here, the function retrieves the current filter category, so its name should be appropriate.

frontend/src/components/organisms/TableFilters/index.jsx
```diff
@@ -41,10 +67,16 @@ const TableFilter = (props) => {
...
+    if (!(tableFilters?.length > 0)) {
+      return "";
+    }
+    const filterResults = tableFilters.filter((item) => item.name === name);
+    if (filterResults.length == 0) {
```
MVarshini, line 75: Instead of checking the length of the `filterResults` and returning early, it can be handled in the `filter` function with optional chaining.   
`const filterResults = tableFilters.filter((item) => item.name === name)?.[0].value`

frontend/src/components/organisms/TableFilters/index.jsx
```diff
@@ -72,28 +106,33 @@ const TableFilter = (props) => {
...
-                options={filterOptions}
-                onChange={updateSelectedFilter}
-                applyMethod={onOptionsChange}
-                currCategory={category}
-                selected={selectedFilters?.find((i) => i.name === category)}
-                width={"300px"}
-              />
-            </ToolbarItem>
-          </ToolbarContent>
+        {tableFilters != null && filterOptions != null && updateSelectedFilter != null && (
```
MVarshini, line 109: In JavaScript, `length` property should be used to check if an array is empty. 

As `tableFilters` and `filterOptions` are arrays, it can be checked as `tableFilters?.length === 0` which handles `null` and `undefined` cases.

dbutenhof (approved):
But I think Varshini's comments are worth addressing.

frontend/src/components/organisms/TableFilters/index.jsx
jaredoconnell: I switched it to the simplified line that you proposed.

frontend/src/components/organisms/TableFilters/index.jsx
MVarshini: The `tableFilters` array is declared in the initial state of the Redux store, which means it cannot be null unless it was unintentionally skipped when adding new products. Therefore, a null check is unnecessary.

The `updateSelectedFilter` function is not optional, and null-checking it would require checking all other functions in the same block, adding unnecessary overhead. Since PropTypes are already used for type checking, any invalid function passed will trigger a warning.

The filterOptions array is initialized as an empty array in the Redux store and is populated based on the currently selected category from filterData. It may remain empty if there are no options for the selected category. However, simply displaying "No filter present" would prevent users from resetting their selections.

Previously, the filterData structure was:
`[{"key": "field", "values": ["x", "y"]}]`

This was calculated from the API response in the frontend, while `tableFilters` stored all possible table filters in the format:
`[{"key": "field", "name": "Field Name"}]`
These were not duplicates but shared common properties that could be combined.

Now, with the updated filterData structure:
`[{"key": "field", "values": ["x", "y"], "name": "Field Name"}]`

the `tableFilters` array is no longer needed, as the field name can be retrieved directly from filterData.
However, it is still maintained to ensure that none of the filter fields are overlooked when constructing queries in the backend. It can be removed later once the filter PRs are merged with the `revamp`.

So, the check should be `filterData?.length > 0` instead of using additional null checks and displaying a message.

jaredoconnell: > The `tableFilters` array is declared in the initial state of the Redux store, which means it cannot be null unless it was unintentionally skipped when adding new products. Therefore, a null check is unnecessary.

Assumptions that it's present can make it harder to debug when creating a new product. Would you rather this form of a null check or an if statement with a console log, or something else?

> The `updateSelectedFilter` function is not optional, and null-checking it would require checking all other functions in the same block, adding unnecessary overhead. Since PropTypes are already used for type checking, any invalid function passed will trigger a warning.
> 
> The `filterOptions` array is initialized as an empty array in the Redux store and is populated based on the currently selected category from filterData. It may remain empty if there are no options for the selected category. However, simply displaying "No filter present" would prevent users from resetting their selections.
> 
> Previously, the filterData structure was: `[{"key": "field", "values": ["x", "y"]}]`
> 
> This was calculated from the API response in the frontend, while `tableFilters` stored all possible table filters in the format: `[{"key": "field", "name": "Field Name"}]` These were not duplicates but shared common properties that could be combined.
> 
> Now, with the updated filterData structure: `[{"key": "field", "values": ["x", "y"], "name": "Field Name"}]`
> 
> the `tableFilters` array is no longer needed, as the field name can be retrieved directly from filterData. However, it is still maintained to ensure that none of the filter fields are overlooked when constructing queries in the backend. It can be removed later once the filter PRs are merged with the `revamp`.
> 
> So, the check should be `filterData?.length > 0` instead of using additional null checks and displaying a message.

If you don't want anything displayed, would you rather it instead print a message to the console log?

Is there anything preventing us from switching to using `filterData` now in this PR, removing `tableFilters`?

MVarshini: After all filtering PRs are merged, `tableFilters` can be replaced with `filterData`. Currently, filterData lacks the `name` property, which must be retrieved from `tableFilters`.

`PropTypes` validates props and displays warnings in the console if validation fails. Now, it checks only the type, but with `isRequired`, it also checks for missing props.

So, `PropTypes.array.isRequired` and `PropTypes.func.isRequired` ensure that `tableFilters` and `updateSelectedFilter` are passed as props and will display an error in the console if they are missing. This allows us to track any missing or incorrectly typed props in the console during development.

jaredoconnell: I think the best way forward is to have a single source of truth at all times. So even though this PR goes half backwards on that, it makes it use the same data source. Then for a follow-up PR you can switch it to `filterData` in entirety. I won't be assigned to CPT Dashboard anymore starting next week.

I updated PropTypes to set a few of the fields as required. They were not set that way before. In the process I removed the null check.
It will display a message when there are no filters present. I think that's good. Otherwise we should allow null values for projects that may not have any filters.

I think this PR is ready, and you can follow it up with more changes.

MVarshini: That's a good one!

MVarshini (approved):
LGTM!

frontend/src/components/organisms/TableFilters/index.jsx
```diff
@@ -3,6 +3,7 @@ import "react-calendar/dist/Calendar.css";
 import "./index.less";
 
 import {
+  Banner,
```
dbutenhof, line 6: You added the `Banner` import, but I don't see it being used. Ah, well. I see you merged, so someone can clean it up later. I'll bet `ESLint` will catch it.

PR #173, opened 2025-02-25
add call to drop na

mfleader:
## Type of change

- [ ] Refactor
- [ ] New feature
- [X] Bug fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Add two calls to `panda.drop_na()` to remove missing, Not a Number (NaN), values from the given job summary.

- Related Issue #
- Closes #

## Testing
Locally, on my Fedora Linux machine.

I added print statements to `jobFilter()` to show if the `jobConfig.jobIteration` column had any NaNs.

```python
def jobFilter(pdata: dict, data: dict):
    # need at least one record to avoid out of bounds error
    if not pdata or not data:
        return []
    columns = ['uuid','jobConfig.jobIterations']
    pdf = pd.json_normalize(pdata)
    # for jsons without a jobConfig.jobIteration value, json_normalize() 
    # fills in Not a Number (NaN)
    pick_df = pd.DataFrame(pdf, columns=columns)
    print(f"jobFilter pick df {pick_df['jobConfig.jobIterations'].hasnans}")
    # pick_df = pd.DataFrame(pdf, columns=columns).dropna(subset=['jobConfig.jobIterations'])
    iterations = pick_df.iloc[0]['jobConfig.jobIterations']
    df = pd.json_normalize(data)
    # same as above
    # ndf = pd.DataFrame(df, columns=columns).dropna(subset=['jobConfig.jobIterations'])
    ndf = pd.DataFrame(df, columns=columns)
    print(f"jobFilter ndf {ndf['jobConfig.jobIterations'].hasnans}")
    ids_df = ndf.loc[df['jobConfig.jobIterations'] == iterations ]
    return ids_df['uuid'].to_list()
```

Run the backend and frontend.

Open the filter drop down and select `Benchmark`.

Then open a drop down arrow from within a  `virt-density` record.

I added print statements to `jobFilter()` to show show the `uuid` and `jobConfig.jobIteration` columns with and without my change. 

Screenshot From 2025-02-25 14-16-35
Screenshot From 2025-02-25 14-17-26

With my changes there are no missing values in the `jobConfig.jobIteration` column.

Screenshot From 2025-02-25 14-20-31

### DIFF
--- a/backend/app/api/v1/endpoints/ocp/graph.py
+++ b/backend/app/api/v1/endpoints/ocp/graph.py
@@ -245,17 +245,12 @@ def jobFilter(pdata: dict, data: dict):
     pdf = pd.json_normalize(pdata)
     # for jsons without a jobConfig.jobIteration value, json_normalize() 
     # fills in Not a Number (NaN)
-    # pick_df = pd.DataFrame(pdf, columns=columns)
     pick_df = pd.DataFrame(pdf, columns=columns).dropna(subset=['jobConfig.jobIterations'])
-    print(f"jobFilter pick df {pick_df['jobConfig.jobIterations'].hasnans}")    
-    iterations = pick_df.iloc[0]['jobConfig.jobIterations']
     df = pd.json_normalize(data)
     # same as above
     ndf = pd.DataFrame(df, columns=columns).dropna(subset=['jobConfig.jobIterations'])
-    # ndf = pd.DataFrame(df, columns=columns)
-    print(f"jobFilter ndf {ndf['jobConfig.jobIterations'].hasnans}")
-    ids_df = ndf.loc[df['jobConfig.jobIterations'] == iterations ]
-    return ids_df['uuid'].to_list()
+    records_matched = ndf['jobConfig.jobIterations'].isin(pick_df['jobConfig.jobIterations'])
+    return ndf[ records_matched ]['uuid'].unique().tolist()
 
 def burnerFilter(data: dict) :
     #
--- a/backend/app/api/v1/endpoints/ocp/graph.py
+++ b/backend/app/api/v1/endpoints/ocp/graph.py
@@ -243,10 +243,13 @@ def jobFilter(pdata: dict, data: dict):
         return []
     columns = ['uuid','jobConfig.jobIterations']
     pdf = pd.json_normalize(pdata)
-    pick_df = pd.DataFrame(pdf, columns=columns)
+    # for jsons without a jobConfig.jobIteration value, json_normalize() 
+    # fills in Not a Number (NaN)
+    pick_df = pd.DataFrame(pdf, columns=columns).dropna(subset=['jobConfig.jobIterations'])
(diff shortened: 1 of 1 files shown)


backend/app/api/v1/endpoints/ocp/graph.py
```diff
@@ -243,10 +243,13 @@ def jobFilter(pdata: dict, data: dict):
         return []
     columns = ['uuid','jobConfig.jobIterations']
     pdf = pd.json_normalize(pdata)
-    pick_df = pd.DataFrame(pdf, columns=columns)
+    # for jsons without a jobConfig.jobIteration value, json_normalize() 
+    # fills in Not a Number (NaN)
+    pick_df = pd.DataFrame(pdf, columns=columns).dropna(subset=['jobConfig.jobIterations'])
```
dbutenhof, line 248: This seems like overkill and insufficient protection at the same time?

We seem to only care about finding "a" value of `iterations`, and we're arbitrarily grabbing the first row ... which suggests we think they're all the same. You're dropping all the null values here, and then assuming that there's still at least one non-null value to pick. What if there isn't? Would it perhaps be better to use something like a `max` filter, which would return a single value we'd hope would not be null? (And what if *all* the values were missing?)

mfleader: I believe I have addressed this. In order to avoid arbitrarily picking one record I used all of them to match with [pandas.isin()](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.isin.html#pandas.DataFrame.isin).

backend/app/api/v1/endpoints/ocp/graph.py
```diff
@@ -245,17 +245,12 @@ def jobFilter(pdata: dict, data: dict):
...
-    iterations = pick_df.iloc[0]['jobConfig.jobIterations']
     df = pd.json_normalize(data)
     # same as above
     ndf = pd.DataFrame(df, columns=columns).dropna(subset=['jobConfig.jobIterations'])
-    # ndf = pd.DataFrame(df, columns=columns)
-    print(f"jobFilter ndf {ndf['jobConfig.jobIterations'].hasnans}")
-    ids_df = ndf.loc[df['jobConfig.jobIterations'] == iterations ]
-    return ids_df['uuid'].to_list()
+    records_matched = ndf['jobConfig.jobIterations'].isin(pick_df['jobConfig.jobIterations'])
+    return ndf[ records_matched ]['uuid'].unique().tolist()
```
dbutenhof, line 253: Pandas often amazes me. So it knows how to use the `records_match` dict boolean values to select only the `ndf` records with the `.isin` criteria. Huh.

dbutenhof (approved):

backend/app/api/v1/endpoints/ocp/graph.py
mfleader: I'm a `pandas` 🧙

PR #174, opened 2025-02-26
Telco Page Bugs

MVarshini:
## Type of change

- [ ] Refactor
- [ ] New feature
- [x] Bug fix
- [ ] Optimization
- [ ] Documentation Update

## Description

The following bugs are found in the Telco page:

- Missing Filters
- Missing fields in Dynamic column list
- Refactor Metrics component
- Total count is missing in the Metrics tab

[PANDA-768](https://issues.redhat.com/browse/PANDA-768)

- Related Issue #
- Closes #

## Testing

### DIFF
--- a/backend/app/api/v1/commons/constants.py
+++ b/backend/app/api/v1/commons/constants.py
@@ -79,6 +79,7 @@ TELCO_FIELDS_DICT = {
     "endDate": "End Date",
     "jobStatus": "Status",
     "ocpVersion": "Build",
+    "isFormal": "Formal",
 }
 
 FILEDS_DISPLAY_NAMES = {
@@ -107,6 +108,7 @@ FILEDS_DISPLAY_NAMES = {
     "concurrency": "Concurrency",
     "imagePushPulls": "Image Push/Pulls",
     "hitSize": "Hit Size",
+    "isFormal": "Is Formal",
 }
 
 FIELDS_FILTER_DICT = {
--- a/backend/app/api/v1/commons/telco.py
+++ b/backend/app/api/v1/commons/telco.py
@@ -139,9 +139,12 @@ async def getFilterData(
 
     # can be removed once python scripts to determine success or failure are executed directly
     # in the splunk dashboard
-    filterData.append(
-        {"key": "jobStatus", "value": ["success", "failure"], "name": "Status"}
-    )
+    extra_filters = [
+        {"key": "jobStatus", "value": ["success", "failure"], "name": "Status"},
+        {"key": "ciSystem", "value": ["JENKINS"], "name": "CI System"},
+    ]
+    filterData.append(extra_filters)
+
     return {"data": filterData, "total": response["total"]}
 
 
--- a/backend/app/services/splunk.py
+++ b/backend/app/services/splunk.py
@@ -130,7 +130,8 @@ class SplunkService:
                     "values(test_type) AS benchmark, "
                     "values(ocp_version) AS ocpVersion, "
                     "values(ocp_build) AS releaseStream "
-                    "| fields cpu, nodeName, benchmark, ocpVersion, releaseStream"
+                    "values(formal) AS isFormal "
+                    "| fields cpu, nodeName, benchmark, ocpVersion, releaseStream, isFormal, total_records"
                 )
                 # Run Splunk search asynchronously using `oneshot`
                 results_reader = await asyncio.to_thread(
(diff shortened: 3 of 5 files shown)


backend/app/api/v1/endpoints/cpt/cptJobs.py
```diff
@@ -25,161 +26,191 @@
...
+
+        return {
+            "data": (
+                response["data"] if not is_filter else response.get("filterData", [])
+            ),
+            "total": response["total"],
+            "summary": response.get("summary", {}),
+        }
+    except Exception as e:
+        print(f"Error fetching data for {product}: {e}\n{traceback.format_exc()}")
```
dbutenhof, line 72: If this is really a "crazy unexpected error" I don't like just hiding it in a backend printout nobody's likely to see -- and which won't have any useful context if someone does. If this is really an unexpected "internal service error", then I think we should just let the FastAPI fallback handle it. We'll get an HTTP 500 response and the server log will show the traceback to help diagnosis.

backend/app/api/v1/endpoints/cpt/cptJobs.py
```diff
@@ -25,161 +26,191 @@
...
-        "offset": offset + size,
-    }
+        ],
+        return_exceptions=True,
+    )
 
-    if pretty:
-        json_str = json.dumps(response, indent=4)
-        return Response(content=json_str, media_type="application/json")
+    results = [res for res in results if isinstance(res, dict)]
```
dbutenhof, line 114: If anything in the `results` list isn't a `dict`, should we really be throwing it away silently, or should we at least be reporting what it was??

backend/app/api/v1/endpoints/cpt/cptJobs.py
```diff
@@ -25,161 +26,191 @@
...
+                        if val.lower() not in existing_values and val != "":
+                            result_dict[key].append(val)
+                else:
+                    # For numbers (version), just avoid duplicates
+                    result_dict[key] = list(set(result_dict[key] + values))
+            else:
+                result_dict[key] = [s for s in values if str(s).strip()]
+
+    merged_result = [
+        {"key": k, "value": v, "name": FILEDS_DISPLAY_NAMES.get(k, k)}
```
dbutenhof, line 198: Huh ... never noticed this before, but isn't "FILEDS" a typo for "FIELDS"??

backend/app/api/v1/endpoints/cpt/cptJobs.py
```diff
@@ -25,161 +26,191 @@
...
+    #     json_str = json.dumps(response, indent=4)
+    #     return Response(content=json_str, media_type="application/json")
 
-    Returns:
-        int: The number of records that can be fetched, which may be less than or equal to requested_size.
-    """
-    available_data = total_count - offset  # Data available from the offset
-    return min(available_data, requested_size)
+    # jsonstring = json.dumps(response)
+    # return jsonstring
```
dbutenhof, line 215: Remove or re-instate the commented code. I think this backend "pretty" query parameter is rather silly given that presentation should be a client function, and the `ORJSONResponse` seems cleaner than returning a dumped (serialized) JSON string, but FastAPI can handle just returning a `dict` to represent a JSON response without any help, and that seems like the simplest and most consistent implementation.

backend/app/api/v1/endpoints/cpt/maps/hce.py
```diff
@@ -17,35 +21,104 @@
...
+            item for item in response["filterData"] if item.get("key") in keys_to_keep
+        ]
+
+        return {
+            "total": response.get("total", 0),
+            "filterData": filtered_data,
+            "summary": response.get("summary", {}),
+        }
+    except Exception as e:
+        print(f"Error retrieving filter data: {e}")
```
dbutenhof, line 96: Here's another place -- we're writing a message without any useful context on what happened, and then silently returning `None`.

If this is really unexpected, I'd much prefer that we *don't catch the exception*, and let FastAPI generate an HTTP 500 with a log traceback. If it's expected and should trigger an alternate API behavior, we need to implement that behavior.

dbutenhof (commented):
Not completely finished, but I'm running out of steam today. I'll continue tomorrow!

(I got up to `cpt/maps/quay.py`, which looks to be about halfway down the GitHub scroll bar. Wow ... big PR!)

dbutenhof (approved):
Well that was quite the rebase ... this is a completely different PR than the version I reviewed last week. Some of my previous comments probably still apply to the code base in general, but none apply to changes in this PR!

dbutenhof:
I merged the Quay fixes since nobody else seems to be reviewing ... now you need to resolve conflicts here. 😦

PR #175, opened 2025-03-07
CPT Dashboard Bugs -I

MVarshini:
## Type of change

- [ ] Refactor
- [ ] New feature
- [x] Bug fix
- [ ] Optimization
- [ ] Documentation Update

## Description

This PR resolves the following issues:

- The "Last Updated" time in the header was previously hard-coded; it is now dynamically updated whenever the Jobs API is called.
- The dynamic column checklist was closing after each selection; this behaviour has been fixed.

- Related Issue #
- Closes #

## Testing

### DIFF
--- a/frontend/src/components/molecules/ColumnMenuFilter/index.jsx
+++ b/frontend/src/components/molecules/ColumnMenuFilter/index.jsx
@@ -6,18 +6,17 @@ import {
   SelectList,
   SelectOption,
 } from "@patternfly/react-core";
-import React, { useMemo } from "react";
+import React, { useCallback, useMemo } from "react";
 
 import PropTypes from "prop-types";
 import TableColumnIcon from "@/components/atoms/TableColumnIcon";
-import { uid } from "@/utils/helper";
 import { useSelector } from "react-redux";
 
 /**
  * A component that displays a list of columns available for the project
  * ID type, and allows the user to select or deselect them.
  * Selection changes are communicated through the callback.
- * 
+ *
  * @param {*} props
  *  - type: The project ID type within the CPT Dashboard as used in
  *      the commonActions action file.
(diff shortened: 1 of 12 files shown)


frontend/src/components/organisms/Header/index.jsx
```diff
@@ -43,7 +44,7 @@ const Header = () => {
       <MastheadContent>
         CPT Dashboard
         <div className="last-updated-box">
-          <div>Last Updated: 21/02/2024 13:26:30</div>
+          {updatedTime && <div>Last Updated: {updatedTime}</div>}
```
dbutenhof, line 47: That's interesting. Somehow I always figured that was supposed to be the last time the *dashboard implementation* was updated, and I wondered how the timestamp was set but never got around to looking. So it was just a hardcoded date, and now it's the last time the contents of the current tab were updated?

This doesn't seem all that useful ... although a hardcoded date doesn't help anyone much, either. I suspect we should have a product version that gets displayed on the masthead -- maybe for both the "dashboard" and the "aggregator" (backend) as we move forward, which would be presumably latched into redux from an API call on initial load.

For now ... I'm just kinda curious what the thought process was here?

frontend/src/components/molecules/ColumnMenuFilter/index.jsx
```diff
@@ -32,32 +31,38 @@ const ColumnMenuFilter = (props) => {
...
-    </MenuToggle>
+  const onSelect = useCallback(
+    (_event, value) => {
+      if (activeTableColumns.includes(value)) {
+        props.setColumns(value, false);
+      } else {
+        props.setColumns(value, true);
+      }
+    },
+    [activeTableColumns, props]
```
dbutenhof, line 51: So adding the dependencies fixes the problem? Wow. OK -- the mysteries of React and Javascript. But ... cool.

dbutenhof (approved):
I'm not sure the "timestamp of last tab update" really helps anyone much, but an arbitrary hardcoded date doesn't, either, and I don't think it makes much sense to update that hardcoded date on each PR (which would make it "date of last code change"). I think we need a better "dashboard version you're using" indication, but the old code wasn't helping much. 😆

dbutenhof:
You probably need a new `poetry lock`, maybe a `poetry update` ... apparently there's something in your lock file that the CI doesn't like. You can run `poetry check --lock` manually to confirm.

dbutenhof (approved):

PR #177, opened 2025-03-17
Add and document an e2e test

mfleader:
## Type of change

- [ ] Refactor
- [X] New feature
- [ ] Bug fix
- [ ] Optimization
- [X] Documentation Update

## Description

I added a first draft of a basic end-to-end test, and documented how to execute it. I also added the development dependency [cypress](https://github.com/cypress-io/cypress) as the relevant test framework for end-to-end and component integration tests.

- Related Issue #
- Closes #

- [X] If it is a core feature, I have added thorough tests.

## Testing
See documentation at `frontend/README.md` section **Test**.
Screenshot From 2025-03-17 16-24-31


dbutenhof (approved):
Great start! 🥳 

A few comments:
1. you mention the `ocpperf.toml` without any detail on how to make one (and, more importantly, what assumptions our growing functional test environment might need.) For the backend functional test system I set up a pod with its own Opensearch containing canned data (currently just ILAB) ... we should extend that and the `ocpperf.toml` to cover other tabs. Our functional tests shouldn't be putting a load on real backend databases.
2. For CI deployment, this might involve refactoring my testing pod setup (and Opensearch restore) into a common directory rather than burying it under `backend/tests/functional/setup`. For the backend I've got a `test.sh` that does everything ... adding the UI tests to this script would make sense. (Note that I put the functional tests themselves into a container in the pod; we can do this with the Cypress tests, too, which would keep a testing pod completely isolated from anything outside the pod.)

mfleader:
> Great start! 🥳
> 
> A few comments:
> 
>     1. you mention the `ocpperf.toml` without any detail on how to make one (and, more importantly, what assumptions our growing functional test environment might need.) For the backend functional test system I set up a pod with its own Opensearch containing canned data (currently just ILAB) ... we should extend that and the `ocpperf.toml` to cover other tabs. Our functional tests shouldn't be putting a load on real backend databases.
> 
>     2. For CI deployment, this might involve refactoring my testing pod setup (and Opensearch restore) into a common directory rather than burying it under `backend/tests/functional/setup`. For the backend I've got a `test.sh` that does everything ... adding the UI tests to this script would make sense. (Note that I put the functional tests themselves into a container in the pod; we can do this with the Cypress tests, too, which would keep a testing pod completely isolated from anything outside the pod.)

1) I added a link to the backend configuration in the root README.md.

dbutenhof:
> 1. I added a link to the backend configuration in the root README.md, specified that `poetry` is being used to manage the python dependencies of the backend, and the NodeJS version. What else do you think I need to add?

The problem is that doesn't really help anyone, because the GitHub code doesn't know what values to fill in. Just pointing the reader at the empty template doesn't help them to run the backend ... nothing's going to work. At the same time, we don't want to give the real product Opensearch URLs and credentials; which means that, without a lot of "insider knowledge" running the functional tests is a lot more complicated than this implies!

That's a big advantage of the snapshot mechanism I put together ... the script sets up an Opensearch container and restores a snapshot, and then uses an `ocpperf.toml` pointing at that instance.

At some point (before we can put this into a CI) we need to expand that snapshot to non-ILAB data, and leverage that scripting here. That doesn't need to be done for this PR (it's going to take some work), but the result of omitting it is that nobody can really run this test unless they have access to our internal Opensearch credentials. About the only thing you could realistically say right now is what you've said ... but, implicitly, if you don't have access to a real `ocpperf.toml` and credentials for an Opensearch instance containing the "right" indices, you can't actually do any of this on your own!

dbutenhof (approved):
The update didn't negate my Approve, but I'm re-approving anyway. This is a great first step; although it occurs to me that we've kinda complicated getting to the CI as I've laid down a bunch of infrastructure in my ILAB PRs chain against `main`, while you did this against `revamp`. I'm hoping we can merge `revamp` and at least my early ILAB PRs quickly enough that this isn't a big deal; but you'll probably want to elevate my pod setup out of the `backend/tests/functional` into a top-level `tests` anyway, which means I'll just need to deal with that on my rebase. 😦 

I don't think this PR will impact Varshini's Telco fix, but I'm going to let her go in first. (Hopefully soon.)

dbutenhof:
Oh ... and the Quay fix has a conflict you need to resolve...

mfleader:
> > 1. I added a link to the backend configuration in the root README.md, specified that `poetry` is being used to manage the python dependencies of the backend, and the NodeJS version. What else do you think I need to add?
> 
> The problem is that doesn't really help anyone, because the GitHub code doesn't know what values to fill in. Just pointing the reader at the empty template doesn't help them to run the backend ... nothing's going to work. At the same time, we don't want to give the real product Opensearch URLs and credentials; which means that, without a lot of "insider knowledge" running the functional tests is a lot more complicated than this implies!
> 
> That's a big advantage of the snapshot mechanism I put together ... the script sets up an Opensearch container and restores a snapshot, and then uses an `ocpperf.toml` pointing at that instance.
> 
> At some point (before we can put this into a CI) we need to expand that snapshot to non-ILAB data, and leverage that scripting here. That doesn't need to be done for this PR (it's going to take some work), but the result of omitting it is that nobody can really run this test unless they have access to our internal Opensearch credentials. About the only thing you could realistically say right now is what you've said ... but, implicitly, if you don't have access to a real `ocpperf.toml` and credentials for an Opensearch instance containing the "right" indices, you can't actually do any of this on your own!

That's correct. My assumption is that only folks who know how to configure this software are the only ones currently trying to execute the test suite. I think your points are valid, and your solution is significant, but since that

PR #180, opened 2025-03-19
add gh workflow to check frontend gui

mfleader:
## Type of change

- [ ] Refactor
- [ ] New feature
- [ ] Bug fix
- [ ] Optimization
- [ ] Documentation Update

- Related Issue #
- Closes #

## Testing

### DIFF
--- a/.github/workflows/frontend-check.yaml
+++ b/.github/workflows/frontend-check.yaml
@@ -1,0 +1,45 @@
+name: Cypress tests
+on: push
+jobs:
+  install:
+    runs-on: ubuntu-24.04
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Cypress install
+        uses: cypress-io/github-action@v6
+        with:
+          # Disable running of tests within install job
+          runTests: false
+          build: npm run build
+
+      - name: Save build folder
+        uses: actions/upload-artifact@v4
+        with:
+          name: build
+          if-no-files-found: error
+          path: dist
+          
+  cypress-run:
+    name: Cypress run
+    needs: install
+    runs-on: ubuntu-24.04
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Download the build folder
+        uses: actions/download-artifact@v4
+        with:
+          name: build        
+
+      - name: Cypress run
+        uses: cypress-io/github-action@v6
+        with:
+          start: npm start
+          group: 'UI-Chrome'
+          record: true
+        env:
+          # pass GitHub token to detect new build vs re-run build
+          GITHUB_TOKEN: ${{secrets.GITHUB_TOKEN}}


.github/workflows/frontend-check.yaml
```diff
@@ -0,0 +1,55 @@
...
+        uses: cypress-io/github-action@v6
+        with:
+          start: npm start
+          group: 'UI-Chrome'
+          record: true
+          project: ./frontend
+        env:
+          # pass GitHub token to detect new build vs re-run build
+          GITHUB_TOKEN: ${{secrets.GITHUB_TOKEN}}
+
```
dbutenhof, line 55: You've got a trailing line with what appears to be a single space and no linefeed, so GitHub shows both an empty line *and* the "no trailing linefeed" icon.

.github/workflows/frontend-check.yaml
```diff
@@ -0,0 +1,55 @@
...
+          name: build        
+          path: ./frontend/dist
+
+      - name: Cypress run
+        uses: cypress-io/github-action@v6
+        with:
+          start: npm start
+          group: 'UI-Chrome'
+          record: true
+          project: ./frontend
```
dbutenhof, line 51: So what's actually happening here? You're starting an optimized frontend with no backend, no `ocpperf.toml`? What's going to actually *happen* here? 😦

.github/workflows/frontend-check.yaml
```diff
@@ -0,0 +1,55 @@
+name: Cypress tests
+on: 
+  push:
+    # branches:
+    #   - main
+    # tags:
+    #   - v*
```
dbutenhof, line 7: That's a valid question. I'm not sure we want to trigger on a `push` to *any* branch? But maybe? Do we need to run tests again on defining a tag? I like testing on a merge to `main` (or, in the short term, `revamp`); we don't usually run with additional branches, so this may not matter much.

.github/workflows/frontend-check.yaml
```diff
@@ -0,0 +1,55 @@
...
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Cypress install
+        uses: cypress-io/github-action@v6
+        with:
+          # Disable running of tests within install job
+          runTests: false
+          build: npm run build
```
dbutenhof, line 21: Interesting ... it's complaining about not finding a `.lock` file. Does the `project` tag below not acting as a working directory??

dbutenhof (commented):
I'm really conflicted about all this. Sooner is good, and I hate to massively move code around, and a quick prototype is a fine step. Does this actually work without a backend? In the long run, we need a backend with repeatable known data (e.g., my snapshot model, although that's all buried in the ILAB PR chain now), and it would make sense to use the same test pod between backend and frontend functional tests. But I'm all for doing this step by step in whatever path works ...

mfleader:
> I'm really conflicted about all this. Sooner is good, and I hate to massively move code around, and a quick prototype is a fine step. Does this actually work without a backend? In the long run, we need a backend with repeatable known data (e.g., my snapshot model, although that's all buried in the ILAB PR chain now), and it would make sense to use the same test pod between backend and frontend functional tests. But I'm all for doing this step by step in whatever path works ...

Yes, the one e2e test is simple enough to work without a functioning backend data service, but still complex enough that it would have failed on the most recent bug that was merged.

.github/workflows/frontend-check.yaml
mfleader: I copied this from an arcaflow github workflow, but I'm open to changing it. I wasn't sure what we'd want to do here

dbutenhof: I'm inclined to say we don't need to re-run the CI on tagging; `main` should be kept clean by the PR testing and merge testing... in which case we should remove the commented `tags` phrase entirely.

While I don't think the `branches` phrase here really matters as we don't routinely use long-lived branches, if we're not going to use it we should remove it instead of commenting it. I'd be OK with either removing it or using it with `main` and (temporarily) `revamp`.

dbutenhof (commented):
I don't like merging with commented code, so I think your `push` sub-phrases should "cleaned". Other than that ... while I'm a little uncomfortable about calling it "e2e" when it succeeds without any backend at all (one "e" is completely missing!) this *does* definitely count as a (minimal) "frontend functional test" and, as you pointed out, should have caught the recent bug. (Although we really need to run ESlint in our lint checks, which also would have caught it with less fuss.)

As a first step towards introducing frontend functional testing into the CI, however, this is great!

dbutenhof:
Is this really still "Draft", or are you "Ready for review"?? 😆

dbutenhof (approved):

PR #181, opened 2025-03-20
Bug Fix Divide by Zero

mfleader:
## Type of change

- [ ] Refactor
- [ ] New feature
- [X] Bug fix
- [ ] Optimization
- [ ] Documentation Update

## Description

In the Quay graphs, whenever a data sample has a zero quantity of data points the cpt-dashboard data server throws a divide by zero error. I wrapped the division with a call to `max(datapoints, 1)`. An alternative would be to return zero for these calculations if it is true that [zero buckets means there are no data points which means the data sample is empty](https://github.com/cloud-bulldozer/cpt-dashboard/blob/3a4019d42d3ab954b931e5a1d6d1f3ce407f6d0c/backend/app/api/v1/endpoints/quay/quayGraphs.py#L143).

- Related Issue #
- Closes #

## Testing
I started the application. Clicked the Quay tab. Clicked several Quay tabs and monitored the data server's logs for divide by zero errors. I found none.

### DIFF
--- a/backend/app/api/v1/endpoints/quay/quayGraphs.py
+++ b/backend/app/api/v1/endpoints/quay/quayGraphs.py
@@ -142,7 +142,7 @@ async def parseImageResults(data: dict):
     totals = {"latency": 0.0, "success_count": 0.0, "failure_count": 0.0}
     datapoints = data["aggregations"]["uuid"]["buckets"]
     # if sample size is zero, then our estimates are zero
-    if datapoints == 0:
+    if len(datapoints) == 0:
         return totals
     for each in datapoints:
         safe_add(each, totals, "latency", "latency")
--- a/backend/app/api/v1/endpoints/quay/quayGraphs.py
+++ b/backend/app/api/v1/endpoints/quay/quayGraphs.py
@@ -141,13 +141,16 @@ async def parseApiResults(data: dict):
 async def parseImageResults(data: dict):
     totals = {"latency": 0.0, "success_count": 0.0, "failure_count": 0.0}
     datapoints = data["aggregations"]["uuid"]["buckets"]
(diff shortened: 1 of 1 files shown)


backend/app/api/v1/endpoints/quay/quayGraphs.py
```diff
@@ -145,9 +145,9 @@ async def parseImageResults(data: dict):
         safe_add(each, totals, "latency", "latency")
         safe_add(each, totals, "success_count", "success_count")
         safe_add(each, totals, "failure_count", "failure_count")
-    totals["latency"] /= len(datapoints)
-    totals["success_count"] /= len(datapoints)
-    totals["failure_count"] /= len(datapoints)
+    totals["latency"] /= max(len(datapoints), 1)
+    totals["success_count"] /= max(len(datapoints), 1)
+    totals["failure_count"] /= max(len(datapoints), 1)
```
dbutenhof, line 150: I suppose that'll do the job; but `len(datapoints) == 0`, which causes the problem, means that the `latency`, `success_count`, and `failure_count` are all 0, which is the default initial value, so doing this math to compute `0.0` from `0.0 / 1` seems wasteful.

dbutenhof (approved):
I think `if len(datapoints) == 0: return totals` would be a lot more clear to a reader (and a comment wouldn't hurt). However, it's not a big deal since the loop disappears in that case and the three unnecessary divisions aren't going to take much time compared to the network transfer.

backend/app/api/v1/endpoints/quay/quayGraphs.py
mfleader: I re-implemented as you suggested.

PR #183, opened 2025-04-01
Extend End-to-End Test

mfleader:
## Type of change

- [ ] Refactor
- [X] New feature
- [ ] Bug fix
- [ ] Optimization
- [ ] Documentation Update

## Description

- [x] Extend end-to-end test to cover some behavior of main data table and record toggle with screenshots during the test. The screenshots are uploaded to the job summary as artifacts.
- [x] Add containerfile for backend end-to-end test, and scripts to create an OpenSearch container and seed it with test data from a search database snapshot
- [x] Add step to setup backend service for end-to-end tests in ci
- [X] Update frontend docs to reflect locally setting up and executing the end-to-end tests

- Related Issue #
- Closes #

- [x] If it is a core feature, I have added thorough tests.

## Testing


.github/workflows/frontend-check.yaml
```diff
@@ -1,51 +1,76 @@
...
       - revamp # rm when branch is deleted
   pull_request:
 jobs:
-  install:
-    runs-on: ubuntu-24.04
+  cypress-run-ci:
+    runs-on: ubuntu-latest
     steps:
+      # - name: install podman
+      #   run: apt-get -y install podman
```
dbutenhof, line 13: Let's not check in commented "code" -- at least not without a good reason and an explanatory comment.

backend/tests/db_seed.py
dbutenhof, line 1: This is ... a little ambiguous. For ILAB, where I was only dealing with backend testing, I put the test setup under `backend/tests/functional`; but now that we're merging that with frontend testing, I'm thinking that a root-level `functional/setup` or `tests/functional/setup` might make more sense.

backend/tests/functional/setup/opensearch.yml
```diff
@@ -0,0 +1,17 @@
...
+network.host: 0.0.0.0
+
+# # minimum_master_nodes need to be explicitly set when bound on a public IP
+# # set to 1 to allow single node clusters
+# discovery.zen.minimum_master_nodes: 1
+
+# Setting network.host to a non-loopback address enables the annoying bootstrap checks. "Single-node" mode disables them again.
+# discovery.type: single-node
+
+reindex.remote.allowlist: ["n42-h01-b01-mx750c.rdu3.labs.perfscale.redhat.com:9200"]
```
dbutenhof, line 14: OK, my bug -- but this shouldn't be in the setup we use for testing, it's part of *creating* the test snapshot. 😦  Sigh.

dbutenhof (commented):
I saw this just as I was about to head out to hike, but I couldn't resist a quick pass with some first impressions. More later. 😁

.github/workflows/frontend-check.yaml
```diff
@@ -9,13 +9,15 @@ jobs:
...
       - name: Checkout
         uses: actions/checkout@v4
       - name: run pod composed e2e gui tests
-        run: ./pod-e2e.sh
-        
+        run: |
+          alias podman=docker
+          ./pod-e2e.sh
```
dbutenhof: Why are you running `docker` commands? That shouldn't be necessary ... I was able to use `podman` commands directly, and I'd rather not involve `docker` if we don't have to...

backend/pyproject.toml
```diff
@@ -31,9 +31,11 @@ trio = "^0.18.0"
 uvicorn = "^0.14.0"
 uvloop = "^0.15.2"
 vyper-config = "1.0.0"
+opensearch-py = "^2.8.0"
```
dbutenhof, line 34: I've considered this before, since we could use a much more recent Opensearch; but the backend uses `Elasticsearch` and I don't think it's a great idea to use both...

pod-e2e.sh
```diff
@@ -10,7 +10,7 @@ POD_NAME="pod_e2e"
 
 echo "cleaning up and recreating pod"
 podman pod rm -f ${POD_NAME}
-podman pod create --name=${POD_NAME} --publish 8000
+podman pod create --name=${POD_NAME} --publish 8000:8000
```
dbutenhof: Using the isolated pod structure I created (and assuming you also put your Cypress tests inside a container in the pod), there's no need to expose any ports at all: the pod can be & should be completely isolated.

pod-e2e.sh
```diff
@@ -10,7 +10,7 @@ POD_NAME="pod_e2e"
...
 
 POD_NAME="pod_e2e" ./backend/tests/opensearch_ocp.sh
```
dbutenhof: If you `export POD_NAME="pod_e2e"` then you don't have to redefine the environment variable here to pass it to the script.

Also, I *generated* a random pod name, so that multiple simultaneous pods won't conflict: `export POD_NAME=${POD_NAME:-FUNC${RANDOM}}`. (If you'd prefer `e2e` instead of `FUNC`, that's fine, although "functional test" is a more flexible and general term.)

pod-e2e.sh
```diff
@@ -10,7 +10,7 @@ POD_NAME="pod_e2e"
...
 podman run -d --pod=${POD_NAME} --name=back e2e-backend
```
dbutenhof: And, again, the only thing "functional/e2e test-y" about the backend container is that we map a specific `ocpperf.toml` when we deploy it. There's no need for a separate container image name or build.

```bash
podman run -d ${POD} --name="${POD_NAME}-back" -v "${CPT_CONFIG}:/backend/ocpperf.toml:Z" localhost/backend
```

backend/tests/e2e_backend.containerfile
```diff
@@ -0,0 +1,30 @@
...
+COPY app/ /backend/app/
+COPY tests/ /backend/tests/
+
+RUN cd /backend && \
+    pip install --user poetry && \
+    poetry install    
+
+COPY scripts/ /backend/scripts/
+COPY ocpperf_test.toml /backend/ocpperf.toml
+COPY ocpperf_test.toml /backend/ocpperf_test.toml
```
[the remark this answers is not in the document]
mfleader: At first it was to facilitate easy switching between databases when I was not using a containerized backend, but now it is superfluous. I'll remove it.

backend/tests/db_seed.py
```diff
@@ -0,0 +1,106 @@
...
+        use_ssl = False,
+        verify_certs = False,
+        ssl_assert_hostname = False,
+        ssl_show_warn = False
+    )
+
+
+def get_indices(client_):
+    indices = client_.indices.get_alias(index="*")
+    print(f"found indices {indices}")
```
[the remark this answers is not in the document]
mfleader: I removed it

frontend/e2e_frontend.containerfile
```diff
@@ -0,0 +1,13 @@
...
+# Install app dependencies
+# A wildcard is used to ensure both package.json AND package-lock.json are copied
+# where available (npm@5+)
+COPY package*.json ./
+RUN npm install
+#To bundle your app’s source code inside the Docker image, use the COPY instruction:
+COPY . .
+#Your app binds to port 3000 so you’ll use the EXPOSE instruction to have it mapped by the docker daemon:
+EXPOSE 3000
+CMD ["npm", "run", "cypress:run:ci"]
```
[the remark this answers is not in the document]
mfleader: will do

pod-e2e.sh
```diff
@@ -0,0 +1,21 @@
...
+podman pod rm -f ${POD_NAME}
+podman pod create --name=${POD_NAME} --publish 8000
+
+POD_NAME="pod_e2e" ./backend/tests/opensearch_ocp.sh
+podman run -d --pod=${POD_NAME} --name=back e2e-backend
+echo "seeding db"
+podman exec back poetry run python tests/db_seed.py
+
+podman run --rm -it --pod=${POD_NAME} --name=front e2e-frontend
+exit $?
```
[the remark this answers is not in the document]
mfleader: removed the file

.github/workflows/frontend-check.yaml
mfleader: I assumed `podman` was not available in the ci job environment. I can change it.

backend/pyproject.toml
mfleader: removed `opensearch-py`

pod-e2e.sh
mfleader: the cypress tests can be done in a container, but they don't need to be. I left it this way to be able to reuse this script to setup the dependencies for local development of the e2e tests with the cypress application. I haven't figured out how to make that work when the frontend is containerized.

pod-e2e.sh
mfleader: i like the idea of appending a random string

.github/workflows/frontend-check.yaml
mfleader: removed

backend/ocpperf_test.toml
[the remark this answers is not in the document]
mfleader: yeah, I wasn't sure yet if we wanted one backend config file for the functional tests and the end-to-end tests

backend/tests/e2e_backend.containerfile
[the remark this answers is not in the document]
mfleader: Oh, this is so the test config can live in the repo because [.gitignore ignores ocpperf.toml](https://github.com/cloud-bulldozer/cpt-dashboard/blob/8fce6b649ed99ec7432c501c8cb004ad7c3fc152/.gitignore#L2).

backend/pyproject.toml
mfleader: That makes sense. I also removed any references to `opensearchpy`.

pod-e2e.sh
mfleader: This is also being used as a developer utility to facilitate development of GUI functional tests. To make it support both of the needs of CI and local development, then it would need a cleanup option and logic to remove the pod name with a random string. It's just not necessary at the moment.

pod-e2e.sh
mfleader: This is related to what I was asking the other day about the `functional.containerfile`. I can't override `ENTRYPOINT ["pytest", "-s", "tests/functional"]` to execute my script to seed the database. I wasn't sure how to achieve both its original behavior for pytest and my desired behavior of database seeding.

mfleader: I believe I figured out how to resolve the issues with using the `functional.containerfile`.

backend/tests/e2e_backend.sh
```diff
@@ -0,0 +1,21 @@
...
+
+export POD_NAME="e2e"
+
+echo "cleaning up and recreating pod"
+podman pod rm -f ${POD_NAME}
+podman pod create --name=${POD_NAME} --publish 8000:8000
+
+./backend/tests/opensearch_ocp.sh
+echo "seeding db"
+podman run --rm -it --pod=${POD_NAME} --entrypoint python3 e2e-backend tests/db_seed.py
```
dbutenhof, line 18: Does this need the `-it`? I don't think it should care about `stdin` or a pseudo-tty?? (Not really a big deal, but probably better not to ask for more than we need.)

backend/tests/functional.containerfile
```diff
@@ -0,0 +1,27 @@
...
+    poetry export --without-hashes -f requirements.txt -o requirements.txt && \
+    pip install -U typing-extensions && \
+    pip install --user dash && \
+    pip install --no-cache-dir -r requirements.txt
+
+ADD tests/ /backend/tests/    
+
+ENV SERVER=http://localhost:8000
+
+ENTRYPOINT ["pytest", "-s", "tests/functional"]
```
dbutenhof, line 27: Interesting ... I guess this doesn't matter since you're overriding it; and if you changed it, I'd need to restore it later ... but then, I'll need to "fix" the `--entrypoint` anyway, so maybe it makes more sense just to package your seed script in this container when you build it, and set that as the entrypoint here?

Then again, I wonder if a separate container image for seeding Opensearch instance(s) would be better -- since (as you're doing in your script) we'll want the seed to run inline and first rather than asynchronously, whereas the functional tests can be asynchronous along with the Cypress tests later. I don't think there's a way of untangling all these with this double-duty, and it's easy enough to build another container image instead of using this!

backend/tests/opensearch_ocp.sh
```diff
@@ -0,0 +1,29 @@
...
+
+podman run -d ${POD} --name "${NAME}" \
+    -v "${SETUP}"/functional/setup/ocp_opensearch.yml:/usr/share/opensearch/config/opensearch.yml:z \
+    ${PORTS} \
+    -e "discovery.type=single-node" -e "DISABLE_INSTALL_DEMO_CONFIG=true" \
+    -e "DISABLE_SECURITY_PLUGIN=true" \
+    docker.io/opensearchproject/opensearch:latest
+podman cp ${SETUP}/fixtures/search_db_snapshots/snapshot.tar.gz ${NAME}:/var/tmp/snapshot.tar.gz    
+echo "Unpacking snapshot inside container"
+podman exec "${NAME}" bash -c 'cd /var/tmp ; tar xfz snapshot.tar.gz'
```
dbutenhof, line 28: Again, another seeding alternative would be to add a `-v tests/db_seed.py:/db_seed.py:z` and then follow this with `podman exec "${NAME}" python3 /db_seed.py`. I'm less concerned about "polluting" the Opensearch container(s) because they're effectively "mocks" for our functional testing anyway.

backend/tests/e2e_backend.sh
mfleader: it does not, so i removed it

backend/tests/opensearch_ocp.sh
mfleader: yeah that is an alternative, but I'd like to keep it this way since the functional containerfile has the requisite python dependncies

frontend/e2e_frontend.containerfile
```diff
@@ -0,0 +1,14 @@
...
+
+WORKDIR /usr/src/cpt-dashboard
+# Install app dependencies
+# A wildcard is used to ensure both package.json AND package-lock.json are copied
+# where available (npm@5+)
+COPY package*.json ./
+RUN npm install
+#To bundle your app’s source code inside the Docker image, use the COPY instruction:
+COPY . .
+#Your app binds to port 3000 so you’ll use the EXPOSE instruction to have it mapped by the docker daemon:
```
dbutenhof, line 11: "mapped outside the container" might be better, since we're not using docker.

Having to run functional tests inside a "polluted" frontend context still bothers me, but I guess if there's a way to use it externally against a "clean" frontend, we can figure that out later.

backend/tests/functional.containerfile
mfleader: what does this mean you want to do?

backend/tests/db_seed.py
```diff
@@ -0,0 +1,116 @@
...
+        )
+        response = client_.cluster.health(wait_for_status="yellow")
+        print(f"cluster health: {response['status']}")
+
+        if not seeded:
+           seed_db(client_)
+           seeded = True
+
+        response = client_.indices.get_alias(index="*")
+        print(f'test data index found: {cfg.get("ocp.elasticsearch.indice") in response.keys()}')
```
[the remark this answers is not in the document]
mfleader: yes, but if we import and use `ElasticService` here, then the `functional.containerfile` and `db_seed` will depend on our python package `app` and you previously said you did not want `functional.containerfile` to depend upon our python package

frontend/README.md
```diff
@@ -58,38 +58,36 @@ Then, copy the `build` folder to the proper place on the server for deployment.
...
+npm --prefix frontend run cypress:run:ci
 ```
 
-Execute end-to-end tests within Chrome web browser.
+Alternatively, to interact with the test execution
+
+Start CPT-Dashboard frontend GUI.
 ```shell
-npm run e2e:chrome
+npm --prefix run dev
```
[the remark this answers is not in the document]
mfleader: yep, I added it

frontend/e2e_frontend.containerfile
mfleader: i'll remove those comments

frontend/README.md
```diff
@@ -58,38 +58,36 @@ Then, copy the `build` folder to the proper place on the server for deployment.
...
 - Three shell terminals
-- Current directory is `{project_root}/frontend`
-- A filled out [cpt-dashboard backend  configuration file](/README.md#backend-configuration) named `ocpperf.toml` in `{project_root}/backend`
+- Starting directory is the project root
 - [poetry](https://python-poetry.org/) for managing the backend's Python virtual environment and dependencies
 - NodeJS 22+
 
-Start CPT-Dashboard backend data service.
-
+Start CPT-Dashboard backend data service end-to-end test dependency.
```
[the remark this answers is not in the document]
mfleader: I think both of those are covered. I added headings to highlight them, **Execute E2E Tests** and **Interactive E2E Test Execution**.

mfleader:
https://github.com/cloud-bulldozer/cpt-dashboard/pull/183#discussion_r2026899885

I didn't want to include your [reindex.remote.allowlist](https://github.com/dbutenhof/cpt-dashboard/blob/673e54a11b7d8ebae5eb55a6f605285de6924df2/backend/tests/functional/setup/opensearch.yml#L14) since my opensearch instance didn't need it.

backend/tests/db_seed.py
```diff
@@ -0,0 +1,116 @@
...
+        print(f'at least 5 hits: {ocp_index_hits() >= 5}')
```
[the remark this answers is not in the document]
mfleader: I'll add the `client_` parameter to the function.

backend/tests/e2e_backend.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
+set -e
+
+echo "building backend image"
+podman build -f backend/backend.containerfile --tag backend ./backend
+
+echo "building backend functional test image"
+podman build -f backend/tests/functional.containerfile --tag e2e-backend ./backend
```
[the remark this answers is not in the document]
mfleader: i'll change it for you dave

backend/tests/opensearch_ocp.sh
dbutenhof: Addendum here -- I don't remember exactly how the snapshot tree is organized. It might be possible to merge the two snapshots into one, or separately unpack two (differently named) snapshots onto the same server ... and maybe it's better to just stand up two separate Opensearch servers, although that means changing the port on one. In any case, most of these possibilities (short of rolling everything into one snapshot) imply distinct *names* for the various products, suggesting that we should have `snapshot_ocp.tar.gz` and `snapshot_ilab.tar.gz`

I'm kinda inclined to try to keep our testing pod to one Opensearch instance, which means either consolidating a single snapshot containing all the indices, or being able to unpack/restore each separately -- and using the `prefix` attribute to rename the indices for each `ElasticService` component. This is going to be rather messy to unravel when I rebase my functional tests ... probably the hardest part of merging my ILAB PRs, unfortunately. 😦

frontend/README.md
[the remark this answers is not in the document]
mfleader: There's already a section on **Cloning and Running the Application Locally** with the instructions you're talking about

backend/tests/opensearch_ocp.sh
mfleader: yeah that makes sense, but if you want it to be more like production then we'll need multiple search dbs running

frontend/README.md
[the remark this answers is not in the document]
dbutenhof: The relationships between the top level `README.md`, and the frontend and backend `README.md` files are inconsistent and fragmented. In some sense, I guess the problem here is really more that you added backend deployment into the E2E testing section, which looks odd against the non-testing deployment. (And it's really all "testing" anyway, whether manual or automated, because the product and staging deployments are completely different.) So, fine -- we're not going to make this all clear and consistent here, or maybe ever. (Sigh.)

dbutenhof (commented):
I wish some of the naming and packaging had been better, but we can iterate later. This is OK, but I do want to get REVAMP merged first, which means you'll need to re-target this to the new `main`, so I'm not approving yet. 😁

dbutenhof (approved):

dbutenhof:
Ah ... you need to re-target this to `main` now!

PR #192, opened 2025-04-10
Add Web Server to Deployment and CI

mfleader:
## Type of change

- [ ] Refactor
- [X] New feature
- [ ] Bug fix
- [ ] Optimization
- [ ] Documentation Update

## Description

- [X] add nginx web server configuration file and containerfile for deployment
- [X] modify ci end-to-end tests to use the application deployed with an nginx web server

- Related Issue #
- Closes #

- [X] If it is a core feature, I have added thorough tests.

## Testing


.github/workflows/frontend-check.yaml
```diff
@@ -5,57 +5,28 @@ on:
       - main
   pull_request:
 jobs:
-  install-gui:
-    runs-on: ubuntu-latest
+  e2e-tests:
+    runs-on: ubuntu-24.04
```
dbutenhof, line 9: This needs to be `ubuntu-latest`

dbutenhof (commented):
Having to publish the port alias with the localhost IP bugs me; but unless we can figure an alternative, maybe we're stuck with it and at least for now I suppose it's good enough. 😦 

I'd approve, except I really don't want to drop back to a dependency on the retiring Ubuntu 20.04!

.github/workflows/frontend-check.yaml
mfleader: got it

dbutenhof:
That's weird ... the CI hasn't triggered on this PR for 16 hours, so it's still waiting for the poetry action to check in.

Also, your changing the *name* of the e2e test messed things up a bit ... I changed the branch rules, because I want to require CI pass, but that broke Vishnu's OLS PR which was running with the old test name. I removed the e2e test requirement entirely until this one goes in; but keep in mind that those YAML step names *do* matter!

dbutenhof (dismissed):

backend/tests/e2e_backend.sh
```diff
@@ -11,7 +11,7 @@ export POD_NAME="e2e"
 
 echo "cleaning up and recreating pod"
 podman pod rm -f ${POD_NAME}
-podman pod create --name=${POD_NAME} --publish 8000:8000
+podman pod create --name=${POD_NAME} --publish 127.0.0.1:8000:8000 --publish 127.0.0.1:3000:3000
```
dbutenhof, line 14: Actually, does this mean that the `local-compose.sh` and `pod-compose.sh` launch scripts need to change, too?

dbutenhof: I don't really know what `pod-compose.sh` is supposed to do; it's referencing container images I don't believe exist, and somehow I never even noticed that script before.

For testing it's often useful to set up a local pod, and I think `local-compose.sh` is a good idea. The question is whether it still works with your changes. If it needs to be altered, I'd prefer you do that here rather than defer it to later.

mfleader: I wrote those scripts when I created the project years ago, `pod-compose.sh` can go.

dbutenhof (approved):

PR #194, opened 2025-04-15
Bug Fixes in CPT Dashboard

MVarshini:
## Type of change

- [ ] Refactor
- [ ] New feature
- [x] Bug fix
- [ ] Optimization
- [ ] Documentation Update

## Description

This PR fixes the filtering issue on the Home tab where filters were not applied correctly. It also corrects the 'Others' count on the Summary tab by properly handling missing status values.  It  preserves query parameters during side menu navigation and refactors the sorting and filtering actions for better consistency.

- Related Issue #
- Closes #

### DIFF
--- a/backend/app/api/v1/endpoints/cpt/cptJobs.py
+++ b/backend/app/api/v1/endpoints/cpt/cptJobs.py
@@ -6,6 +6,7 @@ from datetime import datetime, timedelta, date
 from fastapi import APIRouter, Query, Response
 from fastapi.responses import ORJSONResponse
 import pandas as pd
+from urllib.parse import urlencode
 
 from .maps.ocp import ocpMapper, ocpFilter
 from .maps.quay import quayMapper, quayFilter
@@ -102,21 +103,38 @@ async def jobs(
         )
 
     offset, size = normalize_pagination(offset, size)
+    filter_dict = get_dict_from_qs(filter) if filter else {}
+    filter_product = filter_dict.pop("product", None)
+    individual_prod = ["ocp", "telco", "quay"]
+    if filter_product:
+        matched = [p for p in filter_product if p in individual_prod]
+        unmatched = [p for p in filter_product if p not in individual_prod]
+        filter_product = matched + ["hce", "ocm"] if unmatched else matched
+        filter_dict["product"] = unmatched
 
+    prod_list = filter_product if filter_product else list(products.keys())
+    print("jobs")
+    print(prod_list)
+
+    updated_filter_qs = urlencode(filter_dict, doseq=True) if filter else ""
+    print(updated_filter_qs)
     results = await asyncio.gather(
         *[
-            fetch_data_limited(product, start_date, end_date, size, offset, filter)
-            for product in products
+            fetch_data_limited(
+                product, start_date, end_date, size, offset, filter=updated_filter_qs
+            )
+            for product in prod_list
         ],
         return_exceptions=True,
     )
 
     results = [res for res in results if isinstance(res, dict)]
 
-    results_df = pd.concat(
-        [res["data"] for res in results if not res["data"].empty], ignore_index=True
-    )
-
+    non_empty_df = [res["data"] for res in results if not res["data"].empty]
+    if non_empty_df:
+        results_df = pd.concat(non_empty_df, ignore_index=True)
+    else:
+        results_df = pd.DataFrame()
     total_jobs_count = sum(int(res["total"]) for res in results)
 
     response = {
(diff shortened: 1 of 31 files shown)


backend/app/api/v1/endpoints/cpt/cptJobs.py
```diff
@@ -156,20 +172,23 @@ async def filters(
...
     filter_dict = get_dict_from_qs(filter) if filter else {}
+
     filter_product = filter_dict.pop("product", None)
 
-    if filter_product and filter_product not in productsFilter:
-        return Response(
-            content=json.dumps({"detail": "product not supported"}),
-            status_code=400,
-        )
+    individual_prod = ["ocp", "telco", "quay"]
```
dbutenhof, line 178: This is the second time this line has appeared -- `INDIVIDUAL_PROD` (or maybe something more like `PRODUCT_LIST`) `= ("ocp", "telco", "quay")` should be defined either here or maybe in `constants.py` rather than duplicating the literal.

backend/app/api/v1/endpoints/cpt/cptJobs.py
```diff
@@ -102,21 +103,36 @@ async def jobs(
...
+    filter_dict = get_dict_from_qs(filter) if filter else {}
+    filter_product = filter_dict.pop("product", None)
+    individual_prod = ["ocp", "telco", "quay"]
+    if filter_product:
+        matched = [p for p in filter_product if p in individual_prod]
+        unmatched = [p for p in filter_product if p not in individual_prod]
+        filter_product = matched + ["hce", "ocm"] if unmatched else matched
+        filter_dict["product"] = unmatched
+
+    prod_list = filter_product if filter_product else list(products.keys())
```
dbutenhof, line 115: I'm losing the logic here -- a comment would help to explain what you're doing. But should this be the `filter_product` (the original input `filter_dict["product"]`), or the *updated* `filter_dict["product"]`?

backend/app/services/search.py
```diff
@@ -246,6 +246,16 @@ async def remove_duplicates(self, all_results):
...
 
+    def get_unique_values(self, all_values):
+        seen = set()
+        values = []
+        for val in all_values:
+            lower_value = str(val).lower()
+            if lower_value not in seen:
+                seen.add(lower_value)
+                values.append(val)
+        return values
```
dbutenhof, line 257: or ...

```python
def get_unique_values(self, all_values):
    return list({str(v).lower() for v in all_values})
```

The whole point of a set is that just putting the values into a set does what your function is doing!

backend/app/services/search.py
```diff
@@ -277,7 +287,10 @@ async def buildFilterData(self, filter, total):
 
             for key, value in refiner.items():
                 field = key
-                values = [bucket["key"] for bucket in value["buckets"]]
+                all_values = [bucket["key"] for bucket in value["buckets"]]
+                # to deduplicate values based on case-insensitive match
+                values = self.get_unique_values(all_values)
+                print(values)
```
dbutenhof, line 293: Don't leave `print` statements like this in the code ... they're not useful to anyone later without context. If something is *likely* to be generally useful for debugging later, you might do something like `print(f"buildFilterData refiner {values}")`, but even that's a bit opaque once you're no longer regularly staring at this code.

Somewhere along the line in my ILAB stuff I've added a Python `Logger`, and we should consider adding log debug/info/warn/error messages as appropriate over time, which will show module and function rather than just a value. For now, we should try to avoid just dumping arbitrary output on the server console that isn't really going to be much help to anyone.

frontend/src/actions/commonActions.js
```diff
@@ -173,3 +91,18 @@ export const getRequestParams = (type) => (dispatch, getState) => {
 
   return params;
 };
+
+export const calculateSummary = (countObj) => {
+  const total = Number(countObj["total"]) || 0;
```
dbutenhof, line 96: Should this be `countObj?.total || 0`?? Or ... if you're sure that `total` exists, why wouldn't it be an integer value already: and if it might *not* exist, the conditional fetch seems safer.

frontend/src/components/organisms/MetricsTab/index.jsx
```diff
@@ -51,6 +50,7 @@ const MetricsTab = (props) => {
     setAppliedFilters(navigation, type);
   };
   const applyOtherFilter = () => {
+    console.log("I'm here");
```
dbutenhof, line 53: ... until I'm not ... 😆 

You should probably delete this now.

backend/app/api/v1/endpoints/cpt/cptJobs.py
MVarshini: Moved the `individual_prod` list to the `constants.py` and renamed it.

backend/app/services/search.py
MVarshini: I wanted to preserve the original casing of the first occurring item , so I came up with this approach

frontend/src/actions/commonActions.js
MVarshini: Yeah right. In some cases, the values in countObj were strings, which led to issues like NaN or unexpected string concatenation during arithmetic operations.

backend/app/api/v1/endpoints/cpt/cptJobs.py
MVarshini: Moved this code to a helper function in `utils.py` with the comments explaining the logic.

backend/app/services/search.py
MVarshini: Got it! I’ll switch to using `Logger` from now on for better logging.

backend/app/api/v1/endpoints/cpt/cptJobs.py
```diff
@@ -156,20 +172,23 @@ async def filters(
...
-    if filter_product and filter_product not in productsFilter:
-        return Response(
-            content=json.dumps({"detail": "product not supported"}),
-            status_code=400,
-        )
+    individual_prod = ["ocp", "telco", "quay"]
+    if filter_product:
+        matched = [p for p in filter_product if p in individual_prod]
+        unmatched = [p for p in filter_product if p not in individual_prod]
+        filter_product = matched + ["hce", "ocm"] if unmatched else matched
```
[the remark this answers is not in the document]
dbutenhof: That's great ... in fact I like your explanation here better than the new comments in the code, but they're good.

mfleader (approved):

PR #197, opened 2025-04-16
Status Filter

MVarshini:
## Type of change

- [ ] Refactor
- [ ] New feature
- [x] Bug fix
- [ ] Optimization
- [ ] Documentation Update

## Description

This PR addresses the filtering issue with the 'Others' card in the Summary tab and corrects a bug related to status filtering in the Telco tab

- Related Issue #
- Closes #

### DIFF
--- a/backend/app/api/v1/commons/constants.py
+++ b/backend/app/api/v1/commons/constants.py
@@ -164,3 +164,4 @@ STANDOUT_PRODUCTS = ["ocp", "telco", "quay"]
 # These products aren't listed by their exact names like "hce" and "ocm",
 # but instead represent grouped categories such as "Developer", "Insights", etc.
 GENERAL_PRODUCTS = ["hce", "ocm"]
+TELCO_STATUS_MAP = {"success": "passed", "failed": "failed", "failure": "failure"}
--- a/backend/app/api/v1/commons/constants.py
+++ b/backend/app/api/v1/commons/constants.py
@@ -129,6 +129,7 @@ FIELDS_FILTER_DICT = {
     "benchmark": "test_type",
     "ocpVersion": "ocp_version",
     "releaseStream": "ocp_build",
+    "jobStatus": "status",
 }
 
 HCE_FIELD_CONSTANT_DICT = {
--- a/backend/app/api/v1/commons/utils.py
+++ b/backend/app/api/v1/commons/utils.py
@@ -160,9 +160,12 @@ def get_dict_from_qs(qs):
 
 def construct_query(filter_dict):
     query_parts = []
+    mapped_status = {"success": "passed", "failed": "failed", "failure": "failure"}
     if isinstance(filter_dict, dict):
         for key, values in filter_dict.items():
             k = constants.FIELDS_FILTER_DICT[key]
+            if k == "status":
+                values = [mapped_status[val] for val in values]
             if len(values) > 1:
                 or_clause = " OR ".join([f'{k}="{value}"' for value in values])
                 query_parts.append(or_clause)
--- a/backend/app/api/v1/commons/utils.py
+++ b/backend/app/api/v1/commons/utils.py
@@ -160,12 +160,13 @@ def get_dict_from_qs(qs):
 
 def construct_query(filter_dict):
     query_parts = []
-    mapped_status = {"success": "passed", "failed": "failed", "failure": "failure"}
+
     if isinstance(filter_dict, dict):
         for key, values in filter_dict.items():
             k = constants.FIELDS_FILTER_DICT[key]
+            # as status values are mapped differenlty for telco
             if k == "status":
-                values = [mapped_status[val] for val in values]
+                values = [constants.TELCO_STATUS_MAP[val] for val in values]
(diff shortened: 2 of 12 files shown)


backend/app/api/v1/commons/utils.py
```diff
@@ -160,9 +160,13 @@ def get_dict_from_qs(qs):
 
 def construct_query(filter_dict):
     query_parts = []
+
     if isinstance(filter_dict, dict):
         for key, values in filter_dict.items():
             k = constants.FIELDS_FILTER_DICT[key]
+            # as status values are mapped differenlty for telco
+            if k == "status":
+                values = [constants.TELCO_STATUS_MAP[val] for val in values]
```
dbutenhof, line 169: Instead of taking an exception if you run into a value that's *not* in the `TELCO_STATUS_MAP` keys, maybe `constants.TELCO_STATUS_MAP.get(val, val)`?

backend/app/api/v1/commons/constants.py
```diff
@@ -163,3 +164,4 @@
 # These products aren't listed by their exact names like "hce" and "ocm",
 # but instead represent grouped categories such as "Developer", "Insights", etc.
 GENERAL_PRODUCTS = ["hce", "ocm"]
+TELCO_STATUS_MAP = {"success": "passed", "failed": "failed", "failure": "failure"}
```
dbutenhof, line 167: So `success` means `passed`, but `failed` means `failed` and `failure` means `failure`? Why is `success` being altered, but not the others?

I'm confused. Maybe a list of the actual `status` values you've observed in Telco data would help.

backend/app/api/v1/endpoints/cpt/cptJobs.py
```diff
@@ -169,6 +169,8 @@ async def filters(
 
     updated_filter_qs = urlencode(filter_dict, doseq=True) if filter else ""
 
+    updated_filter_qs = urlencode(filter_dict, doseq=True) if filter else ""
```
dbutenhof, line 172: Why is this line duplicated before the debug `print`? (And of course we probably don't want to merge with the `print` anyway.)

frontend/src/actions/commonActions.js
```diff
@@ -106,3 +106,15 @@ export const calculateSummary = (countObj) => {
...
   return summary;
 };
+
+export const filterOtherStatus = (filterData) => {
+  const keyWordArr = ["success", "failure"];
+
+  const statusValues =
+    filterData.find((item) => item.key === "jobStatus").value || [];
+
+  const filteredStatus = statusValues.filter(
```
dbutenhof, line 116: So `failed` is "other" rather than counting as a failure??

dbutenhof (commented):
Questions? Confusion? 😆

backend/app/api/v1/commons/constants.py
mfleader: why is `success` a key instead of a value? I thought we needed to transform their status values to `success` or other?

frontend/src/actions/telcoActions.js
```diff
@@ -197,18 +198,14 @@ export const getTelcoSummary = (countObj) => (dispatch) => {
...
-  const data = filteredResults.filter(
-    (item) => !keyWordArr.includes(item.jobStatus?.toLowerCase())
-  );
-  dispatch({
-    type: TYPES.SET_TELCO_FILTERED_DATA,
-    payload: data,
-  });
-  dispatch(tableReCalcValues());
+  const filterData = [...getState().telco.filterData];
+  const summary = getState().telco.summary;
```
mfleader, line 202: I believe we can further factor out the shared implementation details between all the `set<Perf Domain>OtherSummaryFilter()` by using bracket notation to dynamically set the performance domain.

```js
export const setTelcoOtherSummaryFilter = () => (dispatch, getState, perfDomain) => {
  const filterData = [...getState()[perfDomain].filterData];
  const summary = getState()[perfDomain].summary;

  if (summary?.othersCount !== 0) {
    const filteredStatus = filterOtherStatus(filterData);
    dispatch(setSelectedFilter("jobStatus", filteredStatus, false));
    dispatch(setAppliedFilters(perfDomain));
  }
```

hm, `setTelcoAppliedFilters()` and friends would need to be refactored as well.

frontend/src/actions/commonActions.js
MVarshini: No, even "failed" is treated as "failure". This can be referred in the `JOB_STATUS_MAP` in constants.py for the exact mapping. In the response, the `jobStatus` can be  success , failure or other values. So, in the actions filtering out success and failure

backend/app/api/v1/commons/constants.py
MVarshini: In this case, I'm , doing reverse mapping. So when the client sends `jobStatus` as  `success` or `failure`, I'm mapping that back to internal values like "passed", "failed", or "failure" accordingly.

mfleader (approved):

dbutenhof (approved):

PR #199, opened 2025-04-16
Introduce a CPT backend version

dbutenhof:
## Type of change

- [ ] Refactor
- [x] New feature
- [ ] Bug fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Version information is compiled at build time using a Python script, producing a transient JSON file that will be copied into the backend container for access at runtime via `GET /api/version`.

[PANDA-823](https://issues.redhat.com/browse/PANDA-823) CPT aggregator version

## Testing

1. `scripts/start-reload.sh` -- works
2. `./local-compose.sh` -- works
3. `./backend/tests/e2e_backend.sh` -- works
4. GitHub Action -- unfortunately, it seems impractical to test until we merge to main. (That is, PR builds can't use GitHub environment variables or secrets, and I don't want to commit even robot credentials in clear text in my Action workflow, so I can't push a built container anywhere I could pull it to test.) If the `build-push.yml` action fails on merge, I'll need another PR to (hopefully) fix it.


mfleader:
what's the purpose of using the git information? Is it to programmatically incorporate git information into the version?

backend/scripts/version.py
```diff
@@ -0,0 +1,60 @@
...
+    if if_none is not None and len(res) == 0:
+        return if_none
+    if len(res) != 1:
+        raise Exception(f"Command {' '.join(cmd)!r} returned {len(res)} lines: {res}")
+    return res[0]
+
+
+def main():
+    top = Path(getone(["git", "rev-parse", "--show-toplevel"]))
+    backend = top / "backend"
```
mfleader, line 34: I thought `/` was not a supported operator for strings in python. What am I missing here?

mfleader (changes requested):
Why do we need to include all of this information in the version? Is it all supposed to be a part of uniquely identifying any build?

backend/scripts/version.py
dbutenhof: `top` isn't a string, it's a `Path`, and `Path` uses the `/` operator (which conveniently looks like a UNIX directory separator) to construct a new `Path` from the old one with a string tail element; `Path(".") / "foo" / "bar.py"` works out to `./foo/bar.py`.

dbutenhof (commented):
> Why do we need to include all of this information in the version? Is it all supposed to be a part of uniquely identifying any build?

I'm certainly open to arguments that I "went too far"; but I do want to include at least *some* git information because we're going to have multiple development builds between "versions". Using the git commit hash helps to keep them unique. Adding the active *branch* was something of an experiment--if I do two container builds from different local PR branches off `main`, I can tell the servers apart that way. Unfortunately this turns out to be a lot less useful for CI builds as the "branch" isn't distinct or particularly useful (they're headless branches). By the time I figured that out, I didn't feel like backing it out, and it might still be useful for local development.

Including the recent git log history? Well, it was an idea to add context, and I think it's kinda sad (though I haven't bothered to push back) that most of our git commit messages are pretty useless; and if the messages aren't useful, the history probably isn't useful either. My half-formed thought was that I could show the history in a popup from the header version element in the UI; I'm not convinced that's worthwhile ... but I'm also not convinced it's useless. 😆

dbutenhof:
I've been playing with the UI ... I may post a draft PR soon, but FY ...

image

dbutenhof:
Merging this will be an interesting test that we can properly build the container with my version hooks and push it to Quay ... and that ArgoCD actually picks it up and updates the dev deployment.

PR #202, opened 2025-04-21
Fix loading issue on empty data and unify date range logic for consistency

MVarshini:
## Type of change

- [ ] Refactor
- [ ] New feature
- [x] Bug fix
- [ ] Optimization
- [ ] Documentation Update

## Description

This PR fixes the following two issues:

- Resolved an issue where the page kept loading indefinitely when no data was returned, by ensuring the table is properly updated with an empty state.

- Fixed a mismatch between total_jobs in the filter API and the total returned from the jobs API. This was caused by inconsistent date range logic across both endpoints.

- Refactored the date range calculation into a common helper function to ensure consistency and reduce duplication.

- Related Issue #
- Closes #

### DIFF
--- a/backend/app/services/search.py
+++ b/backend/app/services/search.py
@@ -45,6 +45,39 @@ class ElasticService:
             es = AsyncElasticsearch(url, verify_certs=False)
         return es, indice, index_prefix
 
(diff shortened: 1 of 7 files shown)


dbutenhof (approved):
One minor suggestion for future consideration, but I'll resolve it.

PR #203, opened 2025-04-21
add a unit test

mfleader:
## Type of change

- [ ] Refactor
- [X] New feature
- [x] Bug fix
- [ ] Optimization
- [X] Documentation Update

## Description

- [X] Add `vitest` to execute and facilitate writing unit tests.
- [X] Add changes to `vite.config.js` to support unit tests, and bundler optimization whenever we start using a bundler.
- [X] Add a unit test for `formatDate()` in `utils/helper.js`.
- [X] Fixed `formatDate()` to return the correct date string. It was a form of an 'off by one' error. For example, given `1000-9-10` or `1000-10-10`, it was returning `1000-09-10`. 
- [X] Document how to execute unit tests.
- [X] Add unit test check to ci.

- Related Issue #
- Closes #

## Testing


frontend/src/utils/__tests__/helper.test.js
```diff
@@ -0,0 +1,23 @@
...
+    expect(formatDate("1000-10-10")).toBe("1000-10-10");
+  });
+  it("month string is length 1", () => {
+    expect(formatDate("1000-9-10")).toBe("1000-09-10");
+  });
+  it("day string is length 1", () => {
+    expect(formatDate("1000-10-9")).toBe("1000-10-09");
+  });
+  it("year string is length 2", () => {
+    expect(formatDate("01-10-09")).toBe("2009-01-10");
```
dbutenhof, line 16: Does "01/10/09" work? Or "1000/09/10"? They *should*, and since you've tried various combinations with dashes it probably makes sense to try. With and without padding? It's not a complicated function, so it's not too hard to be reasonably exhaustive.

mfleader: I think this is a good idea for a test case in a component or end-to-end integration or functional test as a part of our contract to the user. However, since the Javascript `Date()` constructor is responsible for parsing the input date string, I do not think it is appropriate for our unit test to reproduce tests that likely already exist for `Date()`.

dbutenhof: The argument isn't entirely invalid; however if you really intending to follow that philosophy then you've already got more test cases than necessary. By that logic I would expect to see one happy path and one error path, not the multiple "redundant" happy paths you've got.

dbutenhof (dismissed):

frontend/src/utils/__tests__/helper.test.js
mfleader: ah, do you mean like an input of `1000-9-9`?

dbutenhof: If you're implicitly trusting `new Date`'s parser, I don't think it really matters what string you use -- one string that is correctly parsed and one that isn't should be fine. That the "correct" one isn't trivially straightforward (e.g., your non-zero-padded example) is probably a good idea, but, again, if there's any actual question whether any of the possible expected variants will work we're *not* implicitly trusting `Date`.

mfleader: oh! I forgot one of the test cases was for regression, so that we don't reintroduce a potential off by one error

mfleader: should I keep that?

frontend/src/utils/__tests__/helper.test.js
```diff
@@ -0,0 +1,17 @@
...
+
+import { formatDate } from "../helper";
+
+describe("formatDate", () => {
+  it("month string and day string are length 1", () => {
+    expect(formatDate("1000-9-7")).toBe("1000-09-07");
+  });
+  it("ensure day is not off by one", () => {
+    expect(formatDate("1000-09-08")).toBe("1000-09-08");
+  });
```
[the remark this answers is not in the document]
mfleader: There is not. The issue was that `Date` parsed `1000-10-10` with the assumption of UTC midnight (`1000-10-10T00:00:00Z`). Since our original implementation then used local datetime getters (instead of the UTC getters it has now) and the local time is New York Time, or EST, which is still the previous day at 20:00 (UTC-4), hence the `1000-10-09` output.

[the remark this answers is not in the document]
dbutenhof: Yeah; the point is that if we have an important 2nd case, it should be functionally distinct. Given that *any* date would have been subject to the UTC/local "time warp", there should be no value in a second case. And if we add any regression tests for whatever reason, there should probably be a comment explaining what it tests, probably with a reference to the fix--otherwise the reasons will be lost in the mists of time. 😆

dbutenhof (dismissed):
I'm not sure I want to merge anything on `main` until we've got the production deployment working, just in case we have to re-trigger the container build. But aside from all that stuff, this looks good.

dbutenhof (approved):

.github/workflows/frontend-check.yaml
```diff
@@ -20,6 +20,8 @@ jobs:
         run: podman run --pod=${POD_NAME} -d --name=front frontend
       - name: build cypress end-to-end tests
         run: podman build -t e2e-frontend -f frontend/e2e_frontend.containerfile ./frontend
+      - name: run unit tests
+        run: podman run --rm -d e2e-frontend test
```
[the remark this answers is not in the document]
mfleader: I added a `unit-tests` job. Is it appropriate to use the `-it` options to see the output and exit status of the process within the container?

PR #140, opened 2024-12-06
Add a simple unit test framework with `tox` and `pytest`

dbutenhof:
## Type of change

- [ ] Refactor
- [x] New feature
- [ ] Bug fix
- [ ] Optimization
- [ ] Documentation Update

## Description

This uses `black`, `isort` and `flake8` to check code quality, although failure is ignored until we've cleaned it up (a process which begins in PR https://github.com/cloud-bulldozer/cpt-dashboard/pull/139 against the `revamp` branch).

Minimal unit testing is introduced, generating a code coverage report. The text summary is added to the Action summary page, and the more detailed HTML report is stored as an artifact for download.

NOTE: The GitHub Action environment is unhappy with `uvicorn` 0.15 for reasons that aren't entirely obvious (pip can't resolve conflicting indirect dependencies); upgrading to the latest 0.32.x seems to work and hasn't obviously broken anything else.

This is chained from #122 (Crucible service) -> #140 (unit test framework)

## Related Tickets & Documents

[PANDA-673](https://issues.redhat.com/browse/PANDA-673)

Resolves #121 

## Checklist before requesting a review

- [x] I have performed a self-review of my code.
- [ ] If it is a core feature, I have added thorough tests.

## Testing
Tested manually with `tox` CLI in a private branch. If GitHub cooperates, I'll use this draft PR to debug the GitHub Action...

PR #139, opened 2024-12-04
Refactor the Python files for consistent formatting

MVarshini:
## Type of change

- [x] Refactor
- [ ] New feature
- [ ] Bug fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Formatted all the .py files under _backend_ folder as it improves the readability.

## Related Tickets & Documents

- Related Issue # [PANDA-690](https://issues.redhat.com/browse/PANDA-690)

- Closes #

## Checklist before requesting a review

- [ ] I have performed a self-review of my code.
- [ ] If it is a core feature, I have added thorough tests.

## Testing
- Please describe the System Under Test.
- Please provide detailed steps to perform tests related to this code change.
- How were the fix/results from this change verified? Please provide relevant screenshots or results.

PR #205, opened 2025-04-23
Add backend lint & unit test framework.

### LINKED ISSUE #121: [RFE] Testing framework for the dashboard
### Description
As of today our dashboard has naturally grown with a decent amount of changes which displays data related to multiple products like OCP, Quay, Telco and HCP as well. While we have a well grown UI, we do not have a proper testing framework that makes sure if all the components are working as expected which has been a technical debt for a while. So we will need to implement a testing framework to the dashboard.

### Acceptance Criteria
As a contributor/developer one should be able to trigger the testing framework for their changes and generate a coverage report out of it. This should help us to detect breaking changes earlier w.r.t any new additions or updates in the code base.

dbutenhof:
## Type of change

- [ ] Refactor
- [x] New feature
- [ ] Bug fix
- [ ] Optimization
- [ ] Documentation Update

## Description

This uses `black`, `isort` and `flake8` to check code quality, although failure is ignored until we've cleaned it up (which has begin in PR #139 against the `revamp` branch).

Minimal unit testing is introduced, generating a code coverage report. The text summary is added to the Action summary page, and the more detailed HTML report is stored as an artifact for download.

NOTE: The GitHub Action environment is unhappy with `uvicorn` 0.15; upgrading to the latest 0.32.x seems to work and hasn't obviously broken anything else.

[PANDA-673](https://issues.redhat.com/browse/PANDA-673)

Resolves #121

## Testing

Cherry-picked from #140 because rebasing turned out more ugly.

`tox` actions checked locally -- will confirm the GitHub Actions prior to merge.

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -1,0 +1,58 @@
+name: "CodeQL"
+
+on:
+  push:
+    branches: [ main, 'b[0-9].[0-9]+' ]
+  pull_request:
+    # The branches below must be a subset of the branches above
+    branches: [ main, 'b[0-9].[0-9]+' ]
+
+jobs:
+  analyze:
+    name: Analyze
+    runs-on: ubuntu-latest
+    permissions:
+      actions: read
+      contents: read
+      security-events: write
+
+    strategy:
+      fail-fast: false
+      matrix:
+        language: [ 'javascript', 'python' ]
+        # CodeQL supports [ 'cpp', 'csharp', 'go', 'java', 'javascript', 'python', 'ruby' ]
+        # Learn more about CodeQL language support at https://git.io/codeql-language-support
+
+    steps:
+    - name: Checkout repository
+      uses: actions/checkout@v3
+
+    # Initializes the CodeQL tools for scanning.
+    - name: Initialize CodeQL
+      uses: github/codeql-action/init@v2
+      with:
+        languages: ${{ matrix.language }}
+        config-file: ./.github/codeql/${{ matrix.language }}-config.yml
+        # If you wish to specify custom queries, you can do so here or in a config file.
+        # By default, queries listed here will override any specified in a config file.
+        # Prefix the list here with "+" to use these queries and those in the config file.
+        # queries: ./path/to/local/query, your-org/your-repo/queries@main
+
+    # Autobuild attempts to build any compiled languages  (C/C++, C#, or Java).
+    # If this step fails, then you should remove it and run the build manually (see below)
+    #- name: Autobuild
+    #  uses: github/codeql-action/autobuild@v2
+
+    # ℹ️ Command-line programs to run using the OS shell.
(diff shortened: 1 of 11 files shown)


.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,58 @@
...
+    permissions:
+      actions: read
+      contents: read
+      security-events: write
+
+    strategy:
+      fail-fast: false
+      matrix:
+        language: [ 'javascript', 'python' ]
+        # CodeQL supports [ 'cpp', 'csharp', 'go', 'java', 'javascript', 'python', 'ruby' ]
```
mfleader, line 23: i think we can remove this

backend/tests/fake_elastic.py
```diff
@@ -0,0 +1,106 @@
...
+        version: int = 7,
+    ):
+        ver = f"v{version:d}dev"
+        index = f"cdm{ver}-{root_index}"
+        hits = []
+        aggregations = None
+        if hit_list:
+            for d in hit_list:
+                source = d
+                source["cdm"] = {"ver": ver}
```
mfleader, line 38: does this imply that this mock only supports tests against a crucible elasticsearch instance? Was that the intention for this class?

backend/tests/test_crucible.py
```diff
@@ -0,0 +1,265 @@
...
+        assert ("bar", "=") == t._next_token([":", "="])
+        assert ("x", None) == t._next_token([":", "="], optional=True)
+
+    def test_parse_missing(self):
+        """Test exception when a delimiter is missing"""
+
+        t = Parser("foo:bar=x")
+        assert ("foo", ":") == t._next_token([":", "="])
+        assert ("bar", "=") == t._next_token([":", "="])
+        with pytest.raises(HTTPException) as e:
```
mfleader, line 52: so this raises an exception because neither `:` nor `=` were found after `x` in our input string?

mfleader (changes requested):

.github/workflows/codeql-analysis.yml
dbutenhof: I think I picked the comment up from a template somewhere -- yeah, it doesn't really serve much purpose.

backend/tests/fake_elastic.py
dbutenhof: The mock was designed for testing Crucible; while we can and should generalize it, I didn't bother to do that initially. One possibility might be to make an "elastic mock" base class with a "crucible mock" subclass that adds extended behavior.

backend/tests/test_crucible.py
dbutenhof: Actually, `t` has already been advanced beyond the expected fields. Trying to extract the "next" `_next_token` should fail because there *is* no more.

mfleader (approved):
solid work

PR #217, opened 2025-05-02
ILAB UI tab

dbutenhof:
## Type of change

- [ ] Refactor
- [x] New feature
- [ ] Bug fix
- [ ] Optimization
- [ ] Documentation Update

## Description

ILAB UI tab

The original ILAB UI tab was done against main rather than revamp, and needed significant refactoring for consistency. As a "stage of integration", this at least is able to load and display ILAB runs from Opensearch. Some of the filtering and sorting hooks in `ilabActions.js` are no-op-ed out for now as we don't actually use those facilities yet and I'm not going to stray from this original UI support to deal with that here.

[PANDA-857](https://issues.redhat.com/browse/PANDA-857) ILAB UI

## Testing

backend and frontend functional tests, unit tests. A local deployment is able to display the ILAB runs and details.


frontend/src/components/templates/ILab/MetricsDropdown.jsx
```diff
@@ -0,0 +1,86 @@
...
+  const onToggleClick = () => {
+    setIsOpen(!isOpen);
+  };
+  const onSelect = (_event, value) => {
+    console.log("selected", value);
+    const run = value.split("*");
+    //setSelected(run[1].trim());
+    dispatch(setSelectedMetrics(run[0].trim(), run[1].trim()));
+    setIsOpen(false);
+    dispatch(fetchGraphData(run[0].trim(), run[1].trim(), run[2].trim()));
```
MVarshini, line 44: `fetchGraphData` should only be called after all selections are made, focus has moved away, and the dropdown is closed.

frontend/src/components/templates/ILab/MetricsDropdown.jsx
```diff
@@ -0,0 +1,86 @@
...
+          isOpen={isOpen}
+          selected={metrics_selected[item.id]}
+          onSelect={onSelect}
+          onOpenChange={(isOpen) => setIsOpen(isOpen)}
+          toggle={(ref) => toggle1(ref, metrics_selected[item.id])}
+          shouldFocusToggleOnSelect
+        >
+          <SelectList>
+            {getMetricsData(item.id)[0]?.metrics.map((unit) => (
+              <SelectOption
```
MVarshini, line 72: `key` can be item.id instead of calling `uid`, if it's unique

frontend/src/components/templates/ILab/StatusCell.jsx
```diff
@@ -0,0 +1,24 @@
+import {
+  CheckCircleIcon,
+  ExclamationCircleIcon,
+} from "@patternfly/react-icons";
+
+import { Label } from "@patternfly/react-core";
+import Proptype from "prop-types";
+
+const StatusCell = (props) => {
```
MVarshini, line 9: Check for `StatusCell` in `atoms/TableCell.jsx`

frontend/src/components/templates/ILab/MetricsDropdown.jsx
```diff
@@ -0,0 +1,86 @@
...
+              </SelectOption>
+            ))}
+          </SelectList>
+        </Select>
+      )}
+    </>
+  );
+};
+
+export default MetricsSelect;
```
MVarshini, line 86: Consider using the existing `atoms/SelectBox` component here instead of creating a new one

frontend/src/components/templates/ILab/index.jsx
```diff
@@ -0,0 +1,273 @@
...
+        filterData={filterData}
+        start_date={start_date}
+        end_date={end_date}
+        selectedFilters={selectedFilters}
+        updateSelectedFilter={updateSelectedFilter}
+        navigation={navigate}
+        type={"ilab"}
+        showColumnMenu={false}
+      />
+      <Table aria-label="Misc table" isStriped>
```
MVarshini, line 149: It might be helpful to break the table component and its expanded row into smaller, focused components for better readability and maintainability.

frontend/src/components/templates/ILab/index.less
```diff
@@ -0,0 +1,13 @@
+.pf-v5-c-accordion__expandable-content-body {
+    display: block;
```
MVarshini, line 2: This is a shared DOM CSS class, so changing its display property impacts other parts of the app as well. It would be better to scope this rule under a parent class or ID that's unique to this component to avoid unintended side effects.

MVarshini (changes requested):

MVarshini (approved):
Good to ship it!

frontend/src/components/templates/ILab/index.less
MVarshini: Updated the CSS parent's class

PR #222, opened 2025-05-08
transition to podman compose file for dev and testing

mfleader:
## Type of change

- [X] Refactor
- [X] New feature
- [ ] Bug fix
- [ ] Optimization
- [ ] Documentation Update

## Description

- [X] more or less refactor our e2e related scripts into a single podman compose file, `compose.yaml`
- [X] add code to parameterize the python performance data api url of our deployments' at build time and run time, and the target URL for the cypress end-to-end tests
- [X] changed the frontend to use a relative URL without a host in its address to facilitate frontend requests to the api in different environments
- [X] added an `/api/` location to the nginx configuration for reverse proxying, and removed superfluous lines
- [X] added an `/api/` endpoint for vite to reverse proxy in `npm run dev` mode
- [X] updated `local-compose.sh` to use `podman compose`

I'm not happy that the `ocpperf_test.toml`'s `ocp.elasticsearch.url`, and the performance data api host name (used in `frontend/nginx.conf`, `frontend/entrypoint.sh`, etc) now depends upon the name of that service in `compose.yaml`, but I don't know a path around that, and maybe we don't need to solve it right now.

- Related Issue #
- Closes #

- [x] If it is a core feature, I have added thorough tests.

## Testing
Deployed to OpenShift Local on my machine.


compose.yaml
```diff
@@ -0,0 +1,129 @@
+networks:
+  e2e-pod:
+    driver: bridge
+
+services:
+  searchdb-ocp:
```
dbutenhof, line 6: In the vein of "one template to rule them all", ideally we should be setting up both Opensearch instances. (We need frontend ILAB tests and backend OCP tests.)

It also occurs to me that we've got two `snapshot.tar.gz` files in different directories, which is rather unclean. They should be in the same place, with distinct names like `ocp_snapshot.tar.gz` and `ilab_snapshot.tar.gz`. (Sigh.)

Since I'm working now on rebuilding an ILAB snapshot, maybe I'll move it and rename it. Although I think we should have them under `tests/functional`, either in `setup` where I put mine or in a new `tests/functional/fixtures`. 🤔

frontend/README.md
```diff
@@ -95,31 +95,33 @@ npm --prefix frontend run cypress:cp
...
 
-Start frontend GUI server and execute end-to-end tests.
+Generate a CPT Dashboard version file.
+```shell
+./backend/scripts/version.py        
+```
+Execute the end-to-end tests.
 ```shell
-npm --prefix frontend run cypress:run:ci
+./frontend/tests/e2e.sh
```
dbutenhof, line 112: The version generation should be built into the build scripts, not executed manually.

frontend/nginx.conf
```diff
@@ -23,7 +32,33 @@ http {
         location / {
             try_files $uri $uri/ /index.html;
         }
-        
+
+        location /api/ {
```
dbutenhof, line 36: You're still just running `nginx` inside the frontend container, so adding this routing seems ... odd? Do you mean to run NGINX separately in front of both containers? (Which is of course essentially how the real production deployment works under the OpenSearch API router...)

dbutenhof (commented):
Still draft and not running, so I won't worry about detail too much ... but a couple of comments anyway...

compose.yaml
dbutenhof: Yeah, I renamed yours but left it in the same place, and moved mine there in #225; I'd like to make this more consistent, but it can wait.

frontend/nginx.conf
dbutenhof: It "feels" odd that the NGINX inside the frontend container is redirecting local `/api` activity *outside* the container. I suppose that NGINX is happy to do most anything, though; and if it works and doesn't break our other deployment mechanisms, I guess it's OK.

mfleader: in most production deployments there would be a separate ingress pod to handle this part, instead of it being combined with the frontend's static web serving like it is here. We could add an issue to separate these concerns

dbutenhof: NGINX is often cascaded ... our architecture now relies on the frontend NGINX to serve the UI code; it's just that having it do backend routing feels odd. It's more natural to have something like the OpenShift API router do the primary route mapping of `/dashboard` and `/api` to the respective pods, while the frontend NGINX would just handle serving up the local UI artifacts. (And in fact, in the case of the INTLAB OpenShift, I suspect it'll first be going through the INTLAB NGINX proxies to redirect the public-facing `dashboard.app.*` address to the OpenShift router.) And for PBENCH, for example, we also had an NGINX built into the backend, which did the TLS termination and rerouted to a local non-network socket to the server process. "It's NGINX all the way down..."

mfleader: If you'd prefer nginx to be cascaded, I can rework the deployment to separate the web serving and reverse proxy

dbutenhof: Nah, that's really not my issue here -- it's just odd to see the encapsulated frontend NGINX being "overloaded" to route both its own `/dashboard` traffic (serving data within the container image) and `/api` (routing *external* API traffic into the `backend` container).

This is more a discussion than a complaint ... we're interposing NGINX on both  frontend and backend traffic, it's just that the NGINX we're using happens to be packaged with the frontend container image.

Aside from having the frontend container image *assume* the podman network name of the backend, there's nothing much to see here ... it just feels odd.

I guess it does bother me a bit that we're diverging so much from the production deployment; but our development and CI deployments have always differed radically from production so that's nothing new. It's just not moving us any closer; which may just be life. (Though that's why the `podman kube play` sounds so intriguing; it'd be cool if that'd work. But I'm not sure a helm chart counts as "Kubernetes YAML"?)

dbutenhof:
BTW, RE: testing production deployment: quay.io has been read-only most of the day due to an internal configuration issue. No ETA for opening it up, but until then we won't be deploying merges to staging!

https://status.redhat.com/incidents/k7kvfvgfrbdf

frontend/README.md
mfleader: oooooooooooooooh i got you

frontend/nginx.conf
mfleader: I can manually test a deployment with OpenShift Local

mfleader: We could use a single-node kubernetes cluster distribution, but that definitely has its own set of issues.

mfleader: Or, we could change the k8s deployment. The deployment was made that way to reflect the compromise I made in inter-service communication to facilitate agile development in this project's early days of development.

frontend/src/utils/apiConstants.js
```diff
@@ -1,11 +1,4 @@
-export const getUrl = () => {
-  const { hostname, protocol } = window.location;
-  return hostname === "localhost"
-    ? "http://0.0.0.0:8000"
-    : `${protocol}//${hostname}`;
-};
-
-export const BASE_URL = getUrl();
+export const BASE_URL = "";
```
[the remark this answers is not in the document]
dbutenhof: Yes, arguably it could ... however it configures the backend to run against our limited captive Opensearch, which we won't always want for development. `./local-compose.sh` deploys with the local `backend/ocpperf.toml`.

We could add configuration to your deployment to allow that, and it might be worthwhile (outside *this* PR), for in the meantime you're not replacing the need or use of `./local-compose` or a simple `scripts/start-reload.sh` and `npm run dev` ... we need to make sure those work.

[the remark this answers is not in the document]
mfleader: i have a fix for the `npm run dev` usage, but something else broke along the way

frontend/entrypoint.sh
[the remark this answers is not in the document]
dbutenhof: I wish you hadn't created a new test-specific file here; the fact that it joins others that shouldn't be here doesn't really make that more palatable.

frontend/src/utils/apiConstants.js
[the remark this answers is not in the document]
dbutenhof: So `backend/scripts/start-reload.sh` & `npm run dev` works, as well as `./local-compose.sh`??

mfleader:
https://github.com/cloud-bulldozer/cpt-dashboard/pull/222#discussion_r2088904738

`backend/scripts/start-reload.sh` will work if you add in the `poetry run` before the uvicorn command. I can add it now to this pr, if you like

dbutenhof:
> `backend/scripts/start-reload.sh` will work if you add in the `poetry run` before the uvicorn command. I can add it now to this pr, if you like

No; that might be a useful addition, I suppose, but for now I'm happy enough if the script works with `poetry run` externally, or with the environment activated.

> I wasn't planning on addressing local-compose.sh or publish-containers.sh in this pr, but all of their podman calls can be replaced with podman compose stuff.

`publish-containers.sh` is actually a hack to help me deploy standalone containers on the RDU3 host. Once the ILAB code is integrated on `main` we should probably remove it anyway.

`local-compose.sh` is probably "less interesting" than `start-reload.sh`/`npm run dev` since it's an in-between environment. I'd like to be able to continue doing something equivalent to my current `DEVEL=1 backend/tests/functional/setup/test.sh` to stand up a full canned data environment, while `local-compose.sh` runs against the "live" `ocpperf.toml` configuration: but so does `start-reload.sh`, so at worst we could reimplement the arguably convenient `local-compose.sh` by scripting a `start-reload.sh` and `npm run dev`.

> local-compose.sh is mostly easy except for TOP=$(git rev-parse --show-toplevel)

The idea is simply to allow launching the test app regardless of your current directory; I do this a lot with `test.sh`, and I hate scripts that rely on local paths out of an assumed working directory. So, yeah; this is valuable. However, `local-compose.sh` (or some new script) could easily use that to figure out how to run `podman compose up` and we'd be fine.

.github/workflows/frontend-check.yaml
```diff
@@ -32,6 +32,8 @@ jobs:
     steps:
       - name: checkout
         uses: actions/checkout@v4
+      - name: enable podman socket to facilitate podman compose
+        run: systemctl --user enable --now podman.socket
```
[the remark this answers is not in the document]
mfleader: yeah, it's really more of a test to see if we configured the networking and routing correctly; i'll get rid of the spare

local-compose.sh
```diff
@@ -6,16 +6,11 @@
...
 #
 TOP=$(git rev-parse --show-toplevel)
-BACKEND=${TOP}/opt/backend
+BACKEND=${TOP}/backend
 
-CPT_BACKEND_PORT=${CPT_BACKEND_PORT:-8000}
-CPT_FRONTEND_PORT=${CPT_FRONTEND_PORT:-3000}
-CPT_CONFIG=${CPT_CONFIG:-"${BACKEND}/ocpperf.toml"}
-podman rm -f front back
+podman compose --file ${TOP}/compose.yaml down
```
dbutenhof, line 11: Doesn't this generate an error if the "composition" isn't up? Unfortunately, it seems that even `podman compose ps` *fails* if nothing's running, which seems absurd. Maybe a good argument to move away from `compose`, however we choose to do that.

In any case, given that we can't apparently be "smart" about this, I think it might be better just to remove this. The old version failed if the ports were already in use--this can, too.

mfleader: I am not seeing an error

```shell
❯ podman compose down
[+] Running 5/5
 ✔ Container frontend             Removed                                         0.2s 
 ✔ Container db-seed              Removed                                         0.0s 
 ✔ Container backend              Removed                                        10.2s 
 ✔ Container searchdb-ocp         Removed                                         0.4s 
 ✔ Network cpt-dashboard_e2e-pod  Removed                                         0.0s 
❯ echo $?
0
❯ podman compose down
❯ echo $?
0
```

compose.yaml
```diff
@@ -0,0 +1,121 @@
...
+    networks:
+      e2e-pod:
+        aliases: [frontend]
+    ports:
+      - "127.0.0.1:3000:3000"
+
+  frontend-host:
+    <<: *frontend_base
+    profiles: [host]
+    entrypoint: ["nginx", "-g", "daemon off;"]
```
dbutenhof, line 99: Interesting ... this whole `<<:` syntax is new to me and interesting; but in specific right here, why are you overriding the `entrypoint.sh` to drop the DNS check and go directly into NGINX startup? Was the logic not working with `host` networking, for some reason??? 🤔

backend/.dockerignore
dbutenhof, line 1: Seems a bit trivial to exclude the small `.gitignore` file from the build, but OK. However, can we at least have a trailing newline to get rid of those GitHub warning icons? 😆

local-compose.sh
mfleader: do you get that error when you provide it the filepath to `--file`?

compose.yaml
mfleader: oh yeah that was new to me; docker compose v2 used `extends` for inheritance, but that was deprecated in v3, instead they leverage yaml merge, `<<` for composition. 

We need two different network setups for the e2e tests that are isolated on their own network, and interactive development that uses the host's network stack, so I used the merge stuff to stay dry.

mfleader: `host` networking works, but the frontend's [entrypoint.sh relies on dns being able to resolve backend as a host](https://github.com/mfleader/cpt-dashboard/blob/75cfd83e74e81931ac25b997b58376c95a648e83/frontend/entrypoint.sh#L9); [extra_hosts](https://docs.docker.com/reference/compose-file/build/#extra_hosts) makes it possible for the frontend's entrypoint to resolve that dns entry to where it actually is in our local "host" deployment

backend/.dockerignore
mfleader: honestly i have no idea what was happening, but i started getting these weird pipe errors about the .gitignore when i was building the images from my compose file that uses yaml merging for composition; the .dockerignore was the simplest solution

```shell
ERRO[0000] Can't add file /home/mleader/workspace/orgs/cloud-bulldozer/cpt-dashboard/backend/.dockerignore to tar: io: read/write on closed pipe 
ERRO[0000] Can't close tar writer: io: read/write on closed pipe 
ERRO[0000] Can't add file /home/mleader/workspace/orgs/cloud-bulldozer/cpt-dashboard/backend/.dockerignore to tar: io: read/write on closed pipe 
ERRO[0000] Can't close tar writer: io: read/write on closed pipe 
```

dbutenhof:
OK, well *that* didn't go so well. I rebased on your changes, and when I try `./local-compose.sh` I get the following. I wonder what's different on my system?

```
WARNING:podman_compose:missing services [frontend-host,backend-host]
Error: executing /usr/bin/podman-compose --file /home/dbutenho/cpt/deltan/compose.yaml build frontend-host backend-host: exit status 1
Traceback (most recent call last):
  File "/usr/bin/podman-compose", line 8, in <module>
    sys.exit(main())
             ~~~~^^
  File "/usr/lib/python3.13/site-packages/podman_compose.py", line 3709, in main
    asyncio.run(async_main())
    ~~~~~~~~~~~^^^^^^^^^^^^^^
  File "/usr/lib64/python3.13/asyncio/runners.py", line 195, in run
    return runner.run(main)
           ~~~~~~~~~~^^^^^^
  File "/usr/lib64/python3.13/asyncio/runners.py", line 118, in run
    return self._loop.run_until_complete(task)
           ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^
  File "/usr/lib64/python3.13/asyncio/base_events.py", line 719, in run_until_complete
    return future.result()
           ~~~~~~~~~~~~~^^
  File "/usr/lib/python3.13/site-packages/podman_compose.py", line 3705, in async_main
    await podman_compose.run()
  File "/usr/lib/python3.13/site-packages/podman_compose.py", line 1859, in run
    retcode = await cmd(self, args)
              ^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.13/site-packages/podman_compose.py", line 2661, in compose_up
    excluded = get_excluded(compose, args)
  File "/usr/lib/python3.13/site-packages/podman_compose.py", line 2601, in get_excluded
    excluded -= set(x.name for x in compose.services[service]["_deps"])
                                    ~~~~~~~~~~~~~~~~^^^^^^^^^
KeyError: 'frontend-host'
Error: executing /usr/bin/podman-compose --file /home/dbutenho/cpt/deltan/compose.yaml up --detach frontend-host backend-host: exit status 1
dbutenho 08:39 deltan:deltan podman --version
podman version 5.4.2
```

dbutenhof:
Sigh. And, worse, `frontend/tests/e2e.sh` fails, too:

```
+ podman compose --file /home/dbutenho/cpt/deltan/compose.yaml logs --follow e2e-frontend
>>>> Executing external compose provider "/usr/bin/podman-compose". Please see podman-compose(1) for how to disable this message. <<<<

WARNING:podman_compose:missing services [e2e-frontend]
Error: executing /usr/bin/podman-compose --file /home/dbutenho/cpt/deltan/compose.yaml logs --follow e2e-frontend: exit status 1
```

(The "external compose provider" warnings can be removed by setting `PODMAN_COMPOSE_WARNING_LOGS=false`, but I'm not sure what's going on with the "missing services".)

dbutenhof:
Also, the frontend isn't coming up on `dashboard-dev` ... it's looping forever on your "Waiting for backend DNS resolution", so I think that code's not working on the production deployment.

PR #224, opened 2025-05-12
Include Sorting in Telco Table

MVarshini:
PANDA-872

Include Sorting in Telco Table

## Type of change

- [ ] Refactor
- [ ] New feature
- [x] Bug fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Sorting has now been included in the Telco Table.

**Frontend:**

-   sortParam built like fieldName:direction.
-  Sort parameter included in the API request 

**Backend (Python):**

-  sortParam parsed into a list/dict like [{ "nodeName": { "order": "asc" } }].
-  SPL query includes | sort -/+ fieldName.

- Related Issue #  [PANDA-872](https://issues.redhat.com/browse/PANDA-872)
- Closes #

## Testing
image


backend/app/services/splunk.py
```diff
@@ -36,16 +36,34 @@ def __init__(self, configpath="", index=""):
...
             return None
 
-    def build_search_query(self, searchList=""):
+    def build_search_query(self, searchList="", sort=None):
         search_query = f"search index={self.indice} "
 
+        order_symbol = ""
+        sort_field = ""
+
+        if sort and isinstance(sort, list) and sort:
```
dbutenhof, line 45: `if sort [...] and sort:`?? There's no need to test for "truthy" `sort` twice!

(And that's the only concrete "immediate problem" here, and it's not a big deal; but some additional thoughts follow... 😆 )

I also think the `isinstance(sort, list)` is probably excessive unless there are intended to be paths where the value wasn't generated by `build_sort_terms`?? (And, if so, we probably need a more thorough test than just "is list".)

Traditionally the backend has never been anything but a "helper" for the frontend dashboard code; however we really want to move towards treating the backend as a more general "CPT aggregator" service, which will no doubt include more formally documenting and supporting the APIs.

This sort of "if it's exactly what I expect do this otherwise silently ignore it" isn't really a good model for that "new mode", and I'd like to encourage you to think about documentation/support when changing or adding (or reviewing!) code.

What does it mean, for our code or for the user, if `sort` isn't a list?

First, `build_sort_items` raises exceptions (failing the API call with a 400) if the direction is bad, or if there's a bad key value. Otherwise it returns a list of Opensearch-like sort documents. So if we get here and `sort` doesn't have that form, there's an internal server error in processing the request; and we should say that with a reasonable (but brief, as "unexpected errors" are usually, well, unexpected) explanation. (Via something maybe like `raise HTTPException(status.HTTP_500_INTERNAL_SERVER_ERROR, f"Unexpected sort value {sort!r}")`

dbutenhof (dismissed):
The `if sort ... and sort:` bugs me a bit, once I noticed it, but it shouldn't break anything. It'd be great to fix it, though!

backend/app/services/splunk.py
MVarshini: Yeah, that was a typo 🙂 and updated the check to just `if (sort)`. As for `build_sort_items`, totally makes sense and that's already handled. We raise a 500 with a clear message if an unexpected sort format slips through.

dbutenhof (approved):

PR #227, opened 2025-05-13
Enhanced Filter Usability

MVarshini:
Enhance the filter usability

## Type of change

- [ ] Refactor
- [ ] New feature
- [x] Bug fix
- [ ] Optimization
- [ ] Documentation Update

## Description

This PR addresses the following issues for effortless filtering experience 

- Filters are applied and if the API is failed or no result is returned, the UI shows an empty state with an appropriate message.
-  Avoids unnecessary API calls by preventing triggers when filters remain unchanged.
-  Fixed an issue where the Home tab displayed more entries than the selected page size.

- Related Issue #
- Closes #

## Testing
image

### DIFF
--- a/frontend/cypress/e2e/home.cy.js
+++ b/frontend/cypress/e2e/home.cy.js
@@ -28,23 +28,9 @@ describe("basic user journey", () => {
     cy.wait("@fetchSummary").its("response.statusCode").should("eq", 200);
     cy.findByText("Summary").should("be.visible").click({ force: true });
 
-    cy.get("body").then(($body) => {
-      const hasFilter =
-        $body.find('[data-ouia-component-id="data_table_filter"]').length > 0;
-      const hasPagination =
-        $body.find('[data-ouia-component-id="data_table_pagination"]').length >
-        0;
-
-      if (hasFilter) {
-        cy.get('[data-ouia-component-id="data_table_filter"]').should(
-          "be.visible"
-        );
-      }
-      if (hasPagination) {
-        cy.get('[data-ouia-component-id="data_table_pagination"]')
-          .scrollIntoView()
-          .should("be.visible");
-      }
-    });
+    cy.findByTestId("data_table_filter").should("be.visible");
+    cy.findByTestId("data_table_pagination")
+      .scrollIntoView()
+      .should("be.visible");
   });
 });
--- a/frontend/cypress/e2e/home.cy.js
+++ b/frontend/cypress/e2e/home.cy.js
@@ -28,8 +28,11 @@ describe("basic user journey", () => {
     cy.wait("@fetchSummary").its("response.statusCode").should("eq", 200);
     cy.findByText("Summary").should("be.visible").click({ force: true });
 
-    cy.findByTestId("data_table_filter").should("be.visible");
-    cy.findByTestId("data_table_pagination")
+    cy.findByTestId("data_table_filter", { timeout: 30000 })
+      .should("exist")
+      .and("be.visible");
+    cy.findByTestId("data_table_pagination", { timeout: 30000 })
+      .should("exist")
       .scrollIntoView()
       .should("be.visible");
   });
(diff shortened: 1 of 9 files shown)


dbutenhof (approved):
👍🏻

dbutenhof:
Uh oh ... you broke Matt's Cypress e2e test somehow. You get to debug the test. 😆 

image

dbutenhof (changes requested):
We need a test fix to match, apparently.

dbutenhof (commented):
Unfortunately, the e2e tests still didn't work. 😞

frontend/cypress/e2e/home.cy.js
```diff
@@ -9,29 +11,38 @@ describe('basic user journey', () => {
...
-    cy.findByTestId("data_table_filter")
-      .should("be.visible");
+    cy.wait("@fetchData").its("response.statusCode").should("eq", 200);
+    cy.wait("@fetchSummary").its("response.statusCode").should("eq", 200);
+    cy.findByText("Summary").should("be.visible").click({ force: true });
 
-    cy.findByTestId("data_table_pagination")
-      .scrollIntoView()
-      .should("be.visible");
+    cy.get("body").then(($body) => {
```
mfleader, line 27: why move to using low level cypress commands instead of the [cypress testing library plugin](https://testing-library.com/docs/cypress-testing-library/intro/)?

frontend/cypress/e2e/home.cy.js
```diff
@@ -9,29 +11,38 @@ describe('basic user journey', () => {
...
+      const hasFilter =
+        $body.find('[data-ouia-component-id="data_table_filter"]').length > 0;
+      const hasPagination =
+        $body.find('[data-ouia-component-id="data_table_pagination"]').length >
+        0;
+
+      if (hasFilter && hasPagination) {
```
mfleader, line 34: I feel strongly that using [conditional testing](https://docs.cypress.io/app/guides/conditional-testing) is an anti-pattern for testing our application's GUI since we asynchronously fetch data and render our GUI components. I believe the better plan is to control the test fixture inputs, so that we know exactly what to expect in a given test scenario or test case.

frontend/src/components/molecules/EmptyState/index.jsx
```diff
@@ -0,0 +1,30 @@
...
+
+import PropTypes from "prop-types";
+import SearchIcon from "@patternfly/react-icons/dist/esm/icons/search-icon";
+
+const msgToShow = {
+  noData: "No data found for the selected date range.",
+  noFilterData: "No results match the filter criteria.",
+};
+const CustomEmptyState = ({ type }) => (
+  <EmptyState data-testid="custom_empty_state">
```
mfleader, line 16: I believe this should be `ouiaId`, but probably `data-ouia-component-id` since it doesn't inherit from a patternfly component. The corresponding test should also reference that instead of `data-testid`.

mfleader (changes requested):

dbutenhof:
Time to rebase this against main; the last of the stack of ILAB PRs is finally in.

frontend/cypress/e2e/home.cy.js
MVarshini: I agree in general that conditional testing can be considered an anti-pattern, especially when it hides flaky behaviour or masks unexpected changes in the UI.

However, in this case, I’m using this approach because the backend response is asynchronous and can take time, and the cy.get(...) call fails with a timeout if the element hasn’t rendered yet, even the delay is expected.

Using `cy.document().then(...)` allows me to safely check if the element exists without failing the test unnecessarily, and only assert its visibility when it's actually present.

frontend/cypress/e2e/home.cy.js
MVarshini: Moved to `cy.document().then(doc => doc.querySelector(...))` as backend latency causes rendering to be delayed.

 It's closer to the DOM, and bypasses Cypress’s retry behavior but it is used for conditionally rendered elements. This is the only safe way to check if an element might exist without forcing Cypress to fail.

PR #229, opened 2025-05-13
Add metrics pulldown to compare view

dbutenhof:
## Type of change

- [ ] Refactor
- [x] New feature
- [ ] Bug fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Support selection of multiple metrics using the pulldown in the comparison page. The update occurs when the pulldown closes.

To simplify the management of "available metrics" across multiple selected runs, which might have entirely different metrics, the reducer no longer tries to store separate metric selection lists for each run. This also means that the "default" metrics selection remains when adding another comparison run, or expanding another row.
image

[PANDA-865](https://issues.redhat.com/browse/PANDA-865) comparison multi-metric

## Testing

Tested on local deployment


MVarshini (approved):
LGTM!

PR #231, opened 2025-05-14
Improve formatting of delta time in ILAB comparison graph

dbutenhof:
## Type of change

- [ ] Refactor
- [x] New feature
- [ ] Bug fix
- [ ] Optimization
- [ ] Documentation Update

## Description

The Plotly graphing package doesn't directly support a "delta time" type, and in the comparison view we want to use delta time to compare two runs that will generally have different absolute timestamps. (It turns out that the native PatternFly graphing package, Victory, has the same limitation.)

Initially, this just reported numeric delta seconds, but that's unnatural for a reader. This PR adds support for a absolute_relative option which reports the delta times as small absolute timestamps, like 1970-01-01 00:01:00 for 60 seconds, formatting ticks using "%H:%M:%S" ("00:01:00") for readability.

I also made the X axis title appear, which necessitated some refactoring of the layout to avoid overlaying the legend on the axis label; and in the process I moved the "presentation specific" width parameter into the UI and the others into the API so they don't have to be duplicated in the two action calls.
image

[PANDA-866](https://issues.redhat.com/browse/PANDA-866) delta time/ UI fixes

## Testing

Tested using `DEVEL=1 backend/tests/functional/setup/test.sh` deployment.

### DIFF
--- a/frontend/src/actions/ilabActions.js
+++ b/frontend/src/actions/ilabActions.js
@@ -285,15 +285,7 @@ export const fetchGraphData = (uid) => async (dispatch, getState) => {
       graphs,
     });
     if (response.status === 200) {
-      response.data.layout["showlegend"] = true;
-      response.data.layout["responsive"] = "true";
-      response.data.layout["autosize"] = "true";
-      response.data.layout["legend"] = {
-        orientation: "h",
-        xanchor: "left",
-        yanchor: "top",
-        y: -0.1,
-      };
+      response.data.layout["width"] = 1500;
       copyData.push({
         uid,
         data: response.data.data,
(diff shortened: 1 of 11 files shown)


frontend/src/actions/ilabActions.js
```diff
@@ -285,15 +286,7 @@ export const fetchGraphData = (uid) => async (dispatch, getState) => {
...
-      response.data.layout["showlegend"] = true;
-      response.data.layout["responsive"] = "true";
-      response.data.layout["autosize"] = "true";
-      response.data.layout["legend"] = {
-        orientation: "h",
-        xanchor: "left",
-        yanchor: "top",
-        y: -0.1,
-      };
+      // response.data.layout["width"] = 1500;
```
mfleader, line 289: dead code?

mfleader (changes requested):

frontend/src/actions/ilabActions.js
dbutenhof: Undead code has been laid to rest... 🧛🏻

mfleader (approved):

PR #233, opened 2025-05-15
Add a metric label template mechanism

dbutenhof:
## Type of change

- [ ] Refactor
- [x] New feature
- [ ] Bug fix
- [ ] Optimization
- [ ] Documentation Update

## Description

The API makes an effort to create unique metric labels when comparing data across multiple runs, but without human direction the constructed names aren't necessarily ideal to help a human focus on the comparisons they want to make.

This PR adds a UI template mechanism to allow the human to format names that are more useful -- for example, to focus on the differences between software releases or hardware configurations expressed by Crucible tags or params.

For example, a template of `<metric> <iteration>:<period> <tag:accelerator>` identifies metrics between two runs based on the hardware accelerator type used for each.

This introduces a new action to fetch and store the available filters, which exposes the param and tag values currently in use. Usability of the template mechanism will depend on consistent application of specific tags, which we expect to be increasingly supplied automatically by Crucible discovery rather than relying on the (current) ad hoc user definition in endpoint files.
image

[PANDA-868](https://issues.redhat.com/browse/PANDA-868) title templating

## Testing

Tested against local functional test deployment.


MVarshini (approved):
LGTM!

mfleader (approved):

PR #239, opened 2025-05-22
First stage of CDMv9 support

dbutenhof:
## Type of change

- [x] Refactor
- [ ] New feature
- [ ] Bug fix
- [ ] Optimization
- [ ] Documentation Update

## Description

CDMv9 will be more dynamic, as its designed to coexist with V10. With multiple contributing Crucible controllers, we'll want to support interixed V9 and V10 runs. This requires some data restructuring to capture the version of each CDM object rather than assuming all runs are the same.

I've simplified some code by building DTO (data transfer object) classes to abstract differences from most of the code.

This PR is strictly refactoring, with no change in behavior.

[PANDA-898](https://issues.redhat.com/browse/PANDA-898) CDMv9 name pattern

## Testing

- [x] I added thorough DTO unit tests
- [x] Ran both unit and functional ILAB tests
- [x] Ran manual `curl` queries against local server to verify output

### DIFF
--- a/backend/app/services/crucible_readme.md
+++ b/backend/app/services/crucible_readme.md
@@ -66,6 +66,17 @@ store to find information about runs, tags, params, iterations, samples,
 periods, plus various ways to expose and aggregate metric data both for
 primary benchmarks and non-periodic tools.
 
+The Crucible service supports a range of CDM (Common Data Model) versions. An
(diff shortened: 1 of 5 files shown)


MVarshini (approved):
LGTM!

PR #241, opened 2025-05-29
Release Stream Filtering

MVarshini:
Release Stream Filtering

## Type of change

- [ ] Refactor
- [ ] New feature
- [x] Bug fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Release Stream filtering on Home Tab 

- Related Issue #
- Closes #

## Testing
image

### DIFF
--- a/backend/app/api/v1/endpoints/cpt/maps/telco.py
+++ b/backend/app/api/v1/endpoints/cpt/maps/telco.py
@@ -14,8 +14,15 @@ async def telcoMapper(
     start_datetime: date, end_datetime: date, size: int, offset: int, filter: str
 ):
     updated_filter = await get_updated_telco_filter(filter)
+    sort = None
     response = await getData(
-        start_datetime, end_datetime, size, offset, updated_filter, f"telco.splunk"
+        start_datetime,
+        end_datetime,
+        size,
+        offset,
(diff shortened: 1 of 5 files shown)


backend/app/api/v1/endpoints/cpt/maps/telco.py
```diff
@@ -14,8 +14,15 @@ async def telcoMapper(
     start_datetime: date, end_datetime: date, size: int, offset: int, filter: str
 ):
     updated_filter = await get_updated_telco_filter(filter)
+    sort = None
```
dbutenhof, line 17: I won't object to the "constant value variable" here because it does have the advantage of providing a name for each of the `getData` parameters... however a more "Pythonic" way to do that would be to use keyword parameters with `sort=None`. (And using keyword arguments here would have avoided what was apparently a bug that should be breaking the home page Telco mapper right now...)

dbutenhof (approved):

dbutenhof:
Huh. This failed one of my new CDMv9 unit tests. Maybe it needs to be rebased on top of my changes?? Though there shouldn't be any overlap ... strange.

dbutenhof (approved):

PR #250, opened 2025-06-11
Date picker issue

MVarshini:
Date picker on Cypress GUI

## Type of change

- [ ] Refactor
- [ ] New feature
- [x] Bug fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Date picker on Cypress GUI

- Related Issue #
- Closes # [PANDA-895](https://issues.redhat.com/browse/PANDA-895)

## Testing

image

### DIFF
--- a/frontend/src/components/organisms/TableFilters/index.jsx
+++ b/frontend/src/components/organisms/TableFilters/index.jsx
@@ -155,7 +155,7 @@ const TableFilter = (props) => {
           <ToolbarItem>
             <DatePicker
               onChange={(date) =>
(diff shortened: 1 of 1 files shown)


dbutenhof (commented):
The issue was focused on fixing the Cypress test case, which hasn't changed. I thought I remembered Matt commenting something out, though I can't find it now... maybe I need to look back in the history. Should this be changing or adding to the Cypress test cases? If the test case now works, maybe we should probably at least uncomment the Cypress calls that capture the screen shots?

PR #251, opened 2025-06-18
Fully support CDMv9

dbutenhof:
## Type of change

- [x] Refactor
- [x] New feature
- [ ] Bug fix
- [ ] Optimization
- [ ] Documentation Update

## Description

This includes the ability to handle a shared OpenSearch with multiple CDM versions from "v8dev" onward. Queries will span indices so that, for example, runs from `cdmv8dev-run`, `cdm-v9dev-run@2025.05` can be combined naturally.

The `CrucibleService` "object" ID can include information on the index and CDMv9 time-series suffix to optimize APIs taking that ID: for example, when using a run ID to query the iterations or tags for that run, because all the information for a run will have the same time-series suffix.

I've created a new ILAB snapshot, using a massively reworked "clone_ilab.py" tool, which includes runs from both the current private CDMv8 controller and the prototype CDMv9 data from the INTLAB shared OpenSearch, along with many changes to unit and functional tests to accommodate this.

In the process, it was not practical to retain support for the old CDMv7 format. Since both ILAB and Telco Dataplane are using CDMv8, I don't think this will be an issue.

[PANDA-901](https://issues.redhat.com/browse/PANDA-901) mixed version support

## Testing

Local manual testing, plus unit & functional tests.


MVarshini (approved):
LGTM!

PR #254, opened 2025-06-25
ILAB Filter Component Integration

MVarshini:
## Type of change

- [ ] Refactor
- [x] New feature
- [ ] Bug fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Added a filter component for the ILAB tab that enables dynamic filtering of the table data using three chained select boxes as Tag, Name, and Value. Upon selecting a value, the filter is applied, reflected in the URL, and displayed as a removable chip.

**Functionality**
Three Select Boxes:

- Tag (e.g., category/group)
- Name (subcategory or filter key)
- Value (filter value)

On selecting a value:

- The filter is applied to the table.
- The corresponding query is appended to the URL.
- A chip representing the applied filter is shown below the component.

Chips:

- Each chip represents an active filter.
- Clicking the chip's close icon removes the filter and updates the table and URL accordingly.

- Related Issue #
- Closes #

## Testing
image


frontend/src/actions/ilabActions.js
```diff
@@ -101,20 +104,112 @@ export const setIlabDateFilter =
...
+ * Converts the filter string to a grouped filter object
+ * @param {string} appliedFilter
+ * @returns filter object
+ */
+const getGroupedFilters = (appliedFilter) => {
+  const grouped = appliedFilter.split(",").reduce((acc, item) => {
+    const [key, rest] = item.trim().split(":");
+    if (!rest) return acc;
+
+    const value = rest.includes("=") ? rest : rest;
```
dbutenhof, line 178: So `value` is `rest` regardless of whether `rest` includes `=`? 😆 

I don't think the backend parser will *allow* `<section>:<key>` without the `[=|~]<value>`, and I don't think the menu-driven UI needs to worry about `~`.

Apparently you're not interested in the `key=value` here, just in collecting all of the `run` filters, `tag` filters, and `param` filters separately? So, `acc` ends up something like

```json
{
  "tag": ["key=value", "key=value"],
  "param": ["key=value", "key=value"],
}
```

Ah ... this makes it fit in with the "common" 2-level filter mechanisms.

In any case, it seems that `const value = rest;` regardless of whether there's an `=` (which there always should be), so the conditional on this statement seems unnecessary.

Expanding the JSDoc comment to briefly explain what the function is *doing* might help here.

frontend/src/actions/ilabActions.js
```diff
@@ -164,17 +259,39 @@ export const tableReCalcValues = () => (dispatch, getState) => {
 
   dispatch(setIlabPageOptions(page, perPage));
 };
-
+/**
+ * Converts the groubed object to string
```
dbutenhof, line 263: "groubed" => "grouped"

frontend/src/components/templates/ILab/ILabFilters.jsx
```diff
@@ -0,0 +1,173 @@
...
+  const deleteItem = (key, value) => {
+    dispatch(removeIlabAppliedFilters(key, value, navigation));
+  };
+  return (
+    <>
+      {filterData.length > 0 && (
+        <Toolbar
+          id="filter-toolbar"
+          ouiaId="data_table_filter"
+          clearAllFilters={() => console.log("hey")}
```
dbutenhof, line 72: "hey"?? 😁 I suspect there's little if any value to putting debug "bread crumbs" like this into the repo!

dbutenhof (changes requested):
This looks good, although it's "dense" enough that I'm tempted to pull the branch and try it out locally. A few more detailed comments might help, especially since you're "compressing" the three-part ILAB filter terms (section:key=value) to fit into the common "key=value" model. And I had a few other minor comments as well--a trivial typo and a console.log we probably don't need in the repo.

PR #261, opened 2025-07-29
Add cpt dashboard cypress code coverage

MVarshini:
## Type of change

- [ ] Refactor
- [x] New feature
- [ ] Bug fix
- [ ] Optimization
- [ ] Documentation Update

## Description

This PR sets up code coverage collection and reporting for the frontend project using Cypress and c8, integrated into the GitHub Actions workflow.

- Related Issue #
- Closes # [PANDA-848](https://issues.redhat.com/browse/PANDA-848)

## Testing
image


.github/workflows/frontend-check.yaml
```diff
@@ -25,6 +25,36 @@ jobs:
...
+        uses: actions/setup-node@v4
+        with:
+          node-version: '18' 
+          cache: 'npm'
+          cache-dependency-path: ./package-lock.json
+
+      - name: Install dependencies
+        run: npm ci
+      - name: Run Cypress tests and generate coverage
+        run: npm run cypress:run:ci
```
dbutenhof, line 50: It's somewhat interesting that we seem to be generating coverage data based on the functional tests (e2e) rather than on the unit tests as we do on the backend. That's not necessarily bad (I'd kinda like to do both)... but we're already running the frontend functional tests under the `frontend/tests/e2e.sh` script -- doing it twice to collect coverage data seems inefficient. Can we gather the coverage data under `e2e.sh`? 

And does the `cypress:run:ci` script really work? The functional tests require a backend and databases, and it's not clear how all that works in this context. 😕 

(Or maybe it doesn't, and that's the problem we're seeing in the CI run ... ??)

frontend/tests/e2e.sh
```diff
@@ -10,7 +10,15 @@ CYPRESS="${POD_NAME}-cypress"
...
 
 podman build -f ${FRONTEND}/e2e_frontend.containerfile --tag cypress "${FRONTEND}"
-podman run ${POD} --name="${CYPRESS}" localhost/cypress
+
+mkdir -p "${FRONTEND}/coverage"
+
+# podman run --rm \
+#   -v "${FRONTEND}/coverage:/app/coverage" \
+#   --name="${CYPRESS}" \
+#   localhost/cypress
```
dbutenhof, line 19: We don't need the old commented command. (Although the new line is rather long.)

dbutenhof (changes requested):
One section of commented code got left behind, and it'd be nice to clean that up.

frontend/tests/e2e.sh
MVarshini: Updated it!

dbutenhof (approved):

PR #269, opened 2025-08-28
Validate OpenSearch server certificates

dbutenhof:
## Type of change

- [x] Refactor
- [x] New feature
- [ ] Bug fix
- [ ] Optimization
- [ ] Documentation Update

## Description

The CPT dashboard has been accessing most of the OpenSearch datastores using TLS, but with `verify_cert=False`, eliminating most of the security value of having TLS. This change adjusts the backend container build to include the Red Hat CA chain so that we can verify connections to internal datastores, and drops the use of `verify_cert=False`.

This implies a change to the secrets; one of the INTLAB OpenSearch servers (used by HCE) has been using a URL that doesn't have a valid certificate--however there is an alias URL that does validate. I've also considered that there may be a need to expose a TOML configuration option to suppress certificate validation for specific servers: however that's an undesirable configuration from a security point of view, and we can worry about that later if it proves necessary.

The interesting thing is that Red Hat considers the Red Hat CA cert to be internal-only information, which means we don't want to put the URL into GitHub. Therefore, the GitHub Actions now expect a `CA_CERT_PATH` *secret* to be available on the repo, which will be added into the Quay container builds; the cert path isn't needed for the functional tests, which (currently) use local non-TLS unsigned container OpenSearch instances.

This also adds a `run-container.sh` script to build an run a backend container locally in order to test the container build (and CA chain insertion). This requires passing a `CA_CERT_PATH` URL as an environment variable. Under most circumstances, `run-local.sh` is sufficient for testing.

[PANDA-1005](https://issues.redhat.com/browse/PANDA-1005) Red Hat CA chain

## Testing

Local unit and functional tests. The GitHub actions however can't be "debugged" until this PR is merged into main. If something goes wrong, I'll iterate. 🍤  (that's a joke ... typing `:shr` expecting "shrug" includes "shrimp" and "fried shrimp", which is a more interesting emoji, so I started using it to mean "shrug" 🤷🏻 ...)

### DIFF
--- a/backend/tests/unit/test_search.py
+++ b/backend/tests/unit/test_search.py
@@ -135,9 +135,7 @@ class TestElasticService:
         )
 
         mock_es.assert_called_with(
-            "http://localhost:9200",
-            use_ssl=False,
-            http_auth=("testuser", "testpass"),
+            "http://localhost:9200", http_auth=("testuser", "testpass")
         )
         assert indice == "test-index"
         assert prefix == "test-"
--- a/testing/pod_setup.sh
+++ b/testing/pod_setup.sh
@@ -63,7 +63,7 @@ podman pod create --name=${POD_NAME} ${PUBLISH}
 
 echo "Creating version"
 ( cd ${BACKEND}; poetry install; poetry install; poetry run scripts/version.py )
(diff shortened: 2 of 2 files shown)


khandrew-redhat (approved):
Secrets setup and the removal of the certs bypass look good to me. 🍤
Looking forward to trying run-container.sh myself

PR #275, opened 2025-09-30
Back off CA validation on OpenSearch

dbutenhof:
## Type of change

- [ ] Refactor
- [ ] New feature
- [x] Bug fix
- [ ] Optimization
- [ ] Documentation Update

## Description

This problem turned out to be a bit more difficult than I'd hoped, since we build our deployment containers on GitHub Actions with external runners which can't reach the internal Red Hat CA URL. Ultimately the solution would have to be grabbing and installing the CA bundle at container startup (e.g., in the entrypoint); but for now I'm going to declare (temporary) defeat and restore the `verify_certs=False`.

[PANDA-1005](https://issues.redhat.com/browse/PANDA-1005) Red Hat CA chain

## Testing

The real problem is that it's difficult to test GitHub Actions outside the normal triggers, which means we don't see container build problems until merging onto `main`. And, in this case, a local test wouldn't really help since I can't reproduce the GitHub Action Runner environment enough to be sure it makes any difference.

### DIFF
--- a/.github/workflows/build-push.yaml
+++ b/.github/workflows/build-push.yaml
@@ -54,8 +54,6 @@ jobs:
       with:
         image: backend
         context: backend
-        build-args: |
-          CA_CERT_PATH=${{ secrets.CA_CERT_PATH}}
         tags: latest ${{ github.sha }}
         containerfiles: |
           ./backend/backend.containerfile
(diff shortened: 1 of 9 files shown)


MVarshini (approved):
LGTM!

